PR #14, opened 2019-06-10
Feature/165972264 group sdrf and idf files

lingyun1010:


src/test/java/uk/ac/ebi/atlas/download/FileDownloadControllerWIT.java
```diff
@@ -124,6 +123,20 @@ void downloadArchiveForValidExperimentAccession() throws Exception {
...
+
+        this.mockMvc.perform(get(ARCHIVE_DOWNLOAD_URL, EXPERIMENT_ACCESSION)
+                .param("fileType", ExperimentFileType.EXPERIMENT_METADATA.getId()))
+                .andExpect(status().isOk())
+                .andExpect(
+                        header().string(HttpHeaders.CONTENT_DISPOSITION, "attachment;filename=" + expectedArchiveName))
+                .andExpect(content().contentType("application/zip"));
+    }
+
     @Test
```
haideriqbal: It will be good if you can add a test case for `downloadMetadataArchiveForValidExperimentAccessions` as well.

lingyun1010: Thanks for the comment but the metadata download link is in single experiment page. Although it is an archive but it only for a single experiment. In the current context, we do not need a test for multiple experiment accessions.

PR #20, opened 2019-06-24
Bugfix/166870781 reassign row values for marker genes

lingyun1010:


alfonsomunozpomer:
Fixed in https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/39.

PR #31, opened 2019-06-27
Tools jsp with reference to new Single Cell Expression Atlas workflows.

pcm32:
This PR introduces a tool section for single cell home page. In particular, it mentions the inclusion of the new analysis pipelines, pointing to the release tag of them.

### DIFF
--- a/src/main/webapp/WEB-INF/jsp/tiles/views/home.jsp
+++ b/src/main/webapp/WEB-INF/jsp/tiles/views/home.jsp
@@ -16,7 +16,7 @@
 <div class="row column margin-bottom-xlarge expanded">
   <tiles:insertAttribute name="experiments-summary-panel"/>
 </div>
-<div class="row expanded margin-top-large" data-equalizer>
+<div class="row expanded margin-top-large">
   <div class="small-12 medium-12 large-12 columns">
     <tiles:insertAttribute name="tools-box"/>
   </div>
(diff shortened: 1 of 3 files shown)


pcm32:
@irene-papatheodorou please take a look at the wording for the new workflows part. Maybe we should be mentioning CZI funding as well? that many of the tools were jointly developed through CZI Portals call? Do consider that the whole section is a link, so while we can mention things, we cannot link to other things.

src/main/webapp/WEB-INF/jsp/tiles/views/home.jsp
```diff
@@ -16,6 +16,11 @@
 <div class="row column margin-bottom-xlarge expanded">
   <tiles:insertAttribute name="experiments-summary-panel"/>
 </div>
+<div class="row expanded margin-top-large" data-equalizer>
```
alfonsomunozpomer: Since we don’t have several boxes that need to have equal height we can remove `data-equalizer`. Read more about it at https://foundation.zurb.com/sites/docs/equalizer.html.

Also, verify that elements are exactly the same width. I suspect that unless you use `row column`, nesting `small-12 columns` below will add some unwanted extra margin.

lingyun1010: I have deleted `data-equalizer` and tested `row column` and `small-12 columns` issue. The current version keeps the same width as the other panels without extra margin.

PR #40, opened 2019-07-12
Update JsonExperimentsSummaryController.java

hewgreen:
added the following accessions that are both in Atlas and the HCA:

E-GEOD-106540
E-MTAB-6701
E-MTAB-6678
E-GEOD-81547
E-GEOD-93593
E-MTAB-5061
E-ENAD-15

These should all be linked to from the HCA logo.


alfonsomunozpomer:
We now have Tabula Muris as a featured experiment under the CZ Biohub logo. Should we keep it separate or is it better to remove it?

PR #42, opened 2019-08-06
Feature/165531023 alert if missing download files

lingyun1010:


src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +153,58 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+    @RequestMapping(value = "json/experiments/check/zip",
+            method = RequestMethod.GET,
+            produces = "application/json;charset=UTF-8")
+    @ResponseStatus(HttpStatus.OK)
+    @ResponseBody
+    public String
+    checkMultipleExperimentsFileValid(@RequestParam(value = "accession", defaultValue = "") List<String> accessions)
+    {
+        List<Experiment> experiments = new ArrayList<>();
+        for (var accession : accessions) {
```
alfonsomunozpomer: Use streams and `toImmutableList` to build lists in a more declarative way. You’ll get rid of boilerplate code.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +153,58 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+            try {
+                experiments.add(experimentTrader.getPublicExperiment(accession));
+            } catch (Exception e) {
+                LOGGER.debug("Invalid experiment accession: {}", accession);
+            }
+        }
+        List<ExperimentFileType> fileTypeCheckList = new ArrayList<>(Arrays.asList(
```
alfonsomunozpomer: Again, `ImmutableList.of(...)` is your friend here. Also, at the expense of seeming a bit picky, these are not file types, but just files.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +153,58 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+                ExperimentFileType.QUANTIFICATION_RAW,
+                ExperimentFileType.NORMALISED,
+                ExperimentFileType.EXPERIMENT_DESIGN));
+
+       Map<String, List<String>> inValidFilesList = new HashMap<>();
+
+        if (!experiments.isEmpty()) {
```
alfonsomunozpomer: Why are you checking this? What will happen if you remove this check?

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +153,58 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+
+                    else if (fileType != ExperimentFileType.EXPERIMENT_DESIGN) {
+                        var paths = experimentFileLocationService.getFilePathsForArchive(experiment.getAccession(), fileType);
+                        for (var path : paths){
+                            if(path.toFile().exists() == false) {
+                                invalidFiles.add(path.getFileName().toString());
+                            }
+                        }
+                    }
+                }
```
alfonsomunozpomer: Use streams with `filter` and `map`/`flatmap` to do all these in a more straightforward way.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +153,58 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
             zipOutputStream.close();
         }
     }
+
+    @RequestMapping(value = "json/experiments/check/zip",
+            method = RequestMethod.GET,
```
alfonsomunozpomer: If you use `@GetMapping` you can remove the `method` here. You can do so for all endpoints in this file.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +153,58 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
             zipOutputStream.close();
         }
     }
+
+    @RequestMapping(value = "json/experiments/check/zip",
```
alfonsomunozpomer: I think that in this case, a URL that follows the same pattern as above (i.e. `experiments/download/zip`) seems more appropriate. Maybe `experiments/download/zip/check` or `experiments/download/check`?

alfonsomunozpomer (changes requested):
Overall the code would benefit from using streams, which will allow you to get rid of all the emptiness checks. Also, use immutable collections from Guava instead of the regular collections. It helps to reason about the code and reduce the things you need to track when reading it.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: I reformat the code but I think it is file type so I did not change the variable name.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: This is the same logic as line 124. In case invalid experiment accession resulting in empty experiment list. It will trigger errors in the following codes.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: Corrected.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: I correct it as `experiments/download/zip/check`.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: I amend the code with `filter` and `map`.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: I thought we discussed this in `download/zip` implementation if I don't misunderstand. We need to implement this way as we want to `catch` exceptions.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
alfonsomunozpomer: If there were invalid experiment accessions they were skipped when you built `experiments`, so you shouldn’t be concerned about it, right?

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
alfonsomunozpomer: Hmmm... maybe the previous logic was slightly different. With streams you can also catch exceptions, it only gets complicated when you need to throw them out of the lambda block. In this case you can do exactly the same thing without `for`.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +155,57 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+        }
+        ImmutableList<ExperimentFileType> fileTypeCheckList = ImmutableList.of(
+                ExperimentFileType.QUANTIFICATION_RAW,
+                ExperimentFileType.NORMALISED,
+                ExperimentFileType.EXPERIMENT_DESIGN);
+
+       Map<String, List<String>> inValidFilesList = new HashMap<>();
+
+        if (!experiments.isEmpty()) {
+            for (var experiment : experiments) {
```
alfonsomunozpomer: You can change this double `for` loop to use streams instead. It’s an interesting exercise that I encourage you to try and will let you get quite deep with the streams API.

As a hint, build `invalidFilesList` first with one entry per experiment, but only the accessions with missing files will have a non-empty list as a value, and afterwards you can filter out those map entries.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: This check is in case no valid experiment is in the list at all and then we skip the following logic.

PR #43, opened 2019-08-06
Add a URL for current EFO version

lingyun1010:


alfonsomunozpomer (commented):
I’m afraid this wasn’t part of the story! Also, having a link to the EFO OWL file is pretty uninformative to users. In any case we would link to [the EFO site](https://www.ebi.ac.uk/efo/). We can discuss at a later sprint if we want to add links to Ensembl, Ensembl Genomes or the EFO.

PR #46, opened 2019-09-17
Feature/165531023 cherry pick alert missing files

lingyun1010:
Due to the latest updates on single-cell and core repositories, I failed to run/test on the [previous branch](https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/42), so I created a new branch/PR with a unit test for file check controller.

### DIFF
--- a/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
+++ b/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
@@ -154,7 +154,7 @@ public class FileDownloadController extends HtmlExceptionHandlingController {
         }
     }
 
-    @RequestMapping(value = "experiments/check/zip",
+    @RequestMapping(value = "json/experiments/check/zip",
             method = RequestMethod.GET,
             produces = "application/json;charset=UTF-8")
     @ResponseStatus(HttpStatus.OK)
@@ -197,7 +197,12 @@ public class FileDownloadController extends HtmlExceptionHandlingController {
                     }
                 }
 
-                inValidFilesList.put(experiment.getAccession(), invalidFiles);
+                if(!invalidFiles.isEmpty()) {
+                    inValidFilesList.put(experiment.getAccession(), invalidFiles);
+                }
+            }
+            if(!inValidFilesList.isEmpty()){
+                LOGGER.debug("Invalid experiment files: {}", inValidFilesList);
             }
         }
         return GSON.toJson(Collections.singletonMap("invalidFiles", inValidFilesList));
--- a/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
+++ b/src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
@@ -26,6 +26,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Stream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
@@ -135,7 +136,8 @@ public class FileDownloadController extends HtmlExceptionHandlingController {
                 for (var path : paths) {
                     var file = path.toFile();
                     if (file.exists()) {
-                        zipOutputStream.putNextEntry(new ZipEntry(experiment.getAccession() + "/" + file.getName()));
+                        zipOutputStream.putNextEntry(new ZipEntry(
+                                experiment.getAccession() + "/" + file.getName()));
                         FileInputStream fileInputStream = new FileInputStream(file);
 
                         IOUtils.copy(fileInputStream, zipOutputStream);
@@ -151,8 +153,8 @@ public class FileDownloadController extends HtmlExceptionHandlingController {
     @GetMapping(value = "json/experiments/download/zip/check",
                 produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
     @ResponseBody
-    public String validateExperimentsFiles(@RequestParam(value = "accession", defaultValue = "") List<String> accessions)
-    {
+    public String validateExperimentsFiles(
+            @RequestParam(value = "accession", defaultValue = "") List<String> accessions) {
         var experiments = ImmutableSet.<Experiment>builder();
 
         accessions.forEach(accession -> {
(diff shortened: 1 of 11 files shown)


src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +155,60 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+    {
+        List<Experiment> experiments = new ArrayList<>();
+
+        accessions.stream().forEach(accession -> {
+            try{
+                experiments.add(experimentTrader.getPublicExperiment(accession));
+            } catch (Exception e) {
+                LOGGER.debug("Invalid experiment accession: {}", accession);
+            }
+        });
```
alfonsomunozpomer: I’m not really sure that we want to guard ourselves against bad accessions. Since this URL is going to be built and checked from the experiments table component, and is a helper endpoint, we have a reasonable guarantee that it will only contain good accessions, and returning a 404 if a bad accession is provided might not be that bad.

In any case, you be the judge. But whatever choice you make, a couple of points:
- Use `var`.
- Add a blank between try and opening curly brace.
- No need to use `stream()` with `forEach()`.
- Use an `ImmutableList.Builder` or, better yet, since you don’t want duplicates and you’re not accessing elements by index, just go for an `ImmutableSet.Builder`.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +155,60 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+
+        ImmutableList<ExperimentFileType> fileTypeCheckList = ImmutableList.of(
```
alfonsomunozpomer: You can use `var` here, too.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +155,60 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+                ExperimentFileType.QUANTIFICATION_RAW,
+                ExperimentFileType.NORMALISED,
+                ExperimentFileType.EXPERIMENT_DESIGN);
+
+        var filePaths = experiments.stream()
```
alfonsomunozpomer: `filePaths` is of type `Map<String, ImmutableList<List<Path>>>`. Flattening the values so that it is `ImmutableMap<String, ImmutableList<Path>>` will make the filtering second block simpler. You only need a slight change in the mapping step below at L185.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +155,60 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+                                                ImmutableList.of(experimentFileLocationService
+                                                        .getFilePath(experiment.getAccession(), fileType)) :
+                                                experimentFileLocationService
+                                                        .getFilePathsForArchive(experiment.getAccession(), fileType))
+                                        .collect(ImmutableList.toImmutableList())
+
+                        ))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
+        var invalidFilesList = filePaths.keySet().stream()
```
alfonsomunozpomer: No need to iterate over the keys and get the value for each key, you can just do that with `filePaths.entrySet().stream()`.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +155,60 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
             zipOutputStream.close();
         }
     }
+
+    @GetMapping(value = "json/experiments/download/zip/check",
+            produces = "application/json;charset=UTF-8")
+    @ResponseStatus(HttpStatus.OK)
+    @ResponseBody
+    public String
+    checkMultipleExperimentsFileValid(@RequestParam(value = "accession", defaultValue = "") List<String> accessions)
```
alfonsomunozpomer: How about `validateExperimentsFiles`?

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: Corrected.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: Nice one. Renamed.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
lingyun1010: Right. I reformated it using `entrySet()`.

lingyun1010:
Thanks for the comments. I learned a lot about `stream`. I added some tests for experiment accessions parameter, but I can not call `ExperimentBuilder.SingleCellBaselineExperimentBuilder`, so I keep my mock session for now.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +147,56 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+                .collect(toImmutableMap(
+                        Map.Entry::getKey,
+                        experiment -> experiment.getValue()
+                                .stream()
+                                .filter(path -> path.toFile().exists() == false)
+                                .map(path -> path.getFileName().toString())
+                                .collect(toImmutableList())
+                ));
+
+        if(!invalidFilesList.isEmpty()){
```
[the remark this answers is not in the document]
lingyun1010: Installed atlas checkstyle and no problems anymore.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +147,56 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+            }
+        });
+
+        var fileTypeCheckList = ImmutableList.of(
+                ExperimentFileType.QUANTIFICATION_RAW,
+                ExperimentFileType.NORMALISED,
+                ExperimentFileType.EXPERIMENT_DESIGN);
+
+        var filePaths = experiments.build().stream()
+                .distinct()
```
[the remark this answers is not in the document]
lingyun1010: Thanks for the comment. It is very helpful.

src/main/java/uk/ac/ebi/atlas/download/FileDownloadController.java
```diff
@@ -144,4 +147,56 @@ public FileDownloadController(ExperimentFileLocationService experimentFileLocati
...
+                .stream()
+                .collect(toImmutableMap(
+                        Map.Entry::getKey,
+                        experiment -> experiment.getValue()
+                                .stream()
+                                .filter(path -> path.toFile().exists() == false)
+                                .map(path -> path.getFileName().toString())
+                                .collect(toImmutableList())
+                ));
+
```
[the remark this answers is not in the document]
lingyun1010: Very good point! Thanks for the comment. I refactored the code as you suggested as I missed the `flatMap` benefits.

PR #62, opened 2019-10-24
Remove TPM link from non-smart experiments

haideriqbal:
Refer to pivotal story for details:
https://www.pivotaltracker.com/story/show/167757164

Renamed the description of files to make them less confusing and more clear. Also removed the links for TPMs file from non-smart experiments because they exist only in smart experiments.


src/main/java/uk/ac/ebi/atlas/experimentpage/tabs/ExperimentPageContentService.java
```diff
@@ -185,4 +200,10 @@ private JsonObject customContentTab(String tabType, String name, JsonObject prop
         result.add("props", props);
         return result;
     }
+
+    private boolean isSmartExperiment(ImmutableList<String> technologyType) {
+        Stream<Boolean> isSmartExperiment = technologyType.stream()
+                .map(type -> type.toLowerCase().matches(SMART_TECHNOLOGY_TYPE + "-(?:.*)"));
+        return isSmartExperiment.anyMatch(x -> x);
```
lingyun1010: Could you use a more sensible parameter name for `x` please?

haideriqbal:
> Good work and a tiny comment. Please check bamboo test. There is one fail.

It's related to latestAndFeaturedExperiments test. It is passing on my local machine but not on bamboo. And I have not tinkered with any logic related to it. Can it be related to the patch you pushed other day related to an error in featured experiments?

alfonsomunozpomer:
I made two minor modifications to the code and re-wrote the tests to take advantage of AssertJ’s expressiveness. Instead of commenting and request for changes I changed it myself since I think it’s a sort of change that you only see when you’ve used AssertJ enough. However, note that iterating over elements and conditionally make assertions over them is something that can be fully achieved without any looping consructs. Also, note that in your tests, if no elements satisfied your `if` the assertions wouldn’t be run and the tests would pass anyway.

PR #65, opened 2019-11-19
Add param to experiments json endpoint

haideriqbal:
For HCA landing page we need to show anatomograms and intended functionality we want to achieve is when a user clicks on an organism part e.g. lung then all the experiments which contain lung as an organism part are shown. Details regarding the implementation details are in following pivotal ticket:

https://www.pivotaltracker.com/story/show/169594225

### DIFF
--- a/src/main/java/uk/ac/ebi/atlas/experiments/ExperimentsListController.java
+++ b/src/main/java/uk/ac/ebi/atlas/experiments/ExperimentsListController.java
@@ -5,7 +5,6 @@ import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 
-import static com.google.common.base.Strings.isNullOrEmpty;
 import static uk.ac.ebi.atlas.utils.GsonProvider.GSON;
 
 @RestController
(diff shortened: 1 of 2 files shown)


src/main/java/uk/ac/ebi/atlas/experiments/ExperimentsListController.java
```diff
@@ -16,8 +19,9 @@ public ExperimentsListController(ExperimentInfoListService experimentInfoListSer
 
     //Used by experiments table page
     @GetMapping(value = "/json/experiments",
-                produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
-    public String getExperimentsList() {
-        return GSON.toJson(experimentInfoListService.getExperimentsJson());
+            produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
```
alfonsomunozpomer: Align this with `value`, like we do with method parameters.

src/main/java/uk/ac/ebi/atlas/experiments/ExperimentsListController.java
```diff
@@ -16,8 +19,9 @@ public ExperimentsListController(ExperimentInfoListService experimentInfoListSer
...
+    public String getExperimentsList(@RequestParam(defaultValue = "") String organism_part) {
+        return isNullOrEmpty(organism_part) ? GSON.toJson(experimentInfoListService.getExperimentsJson()) :
```
alfonsomunozpomer: We favour Apache Commons `isBlank` instead of `isNullOrEmpty`, just for conciseness. There might be a few stranded uses of `isNullOrEmpty`, though.

src/main/java/uk/ac/ebi/atlas/experiments/ExperimentsListController.java
```diff
@@ -16,8 +19,9 @@ public ExperimentsListController(ExperimentInfoListService experimentInfoListSer
 
     //Used by experiments table page
     @GetMapping(value = "/json/experiments",
-                produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
-    public String getExperimentsList() {
-        return GSON.toJson(experimentInfoListService.getExperimentsJson());
+            produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    public String getExperimentsList(@RequestParam(defaultValue = "") String organism_part) {
```
alfonsomunozpomer: Variables and parameters always in camelCase: `organismPart`.

src/main/java/uk/ac/ebi/atlas/experiments/ExperimentsListController.java
alfonsomunozpomer: Do you need to check if `organismPart` is empty? What would happen if you pass an empty string?

PR #80, opened 2020-02-26
New endpoint for organism_part related experiments

haideriqbal:
Have moved the logic from `atlas-web-core` to single-cell repo. Created new Service, Trader and TraderDao class to keep the code structure in a consistent state. Added test cases for new classes.

Related PR: https://github.com/ebi-gene-expression-group/atlas-web-core/pull/35


src/main/java/uk/ac/ebi/atlas/experiments/ScExperimentDao.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public ImmutableSet<String> fetchExperimentAccessions(String characteristicName, String characteristicValue) {
+        var queryBuilder = new SolrQueryBuilder<SingleCellAnalyticsCollectionProxy>();
+        /* Reason for checking blank here is in case characteristicValue is empty our solrQueryBuilder
+         * builds query like characteristic_value:("\*") which is different from characteristic_value:*
+         * which we need. So to handle that situation if characteristicValue is empty we build query
+         * characteristic_name:("organism_part") as we need all the experiments in case of empty characteristicValue
+         * */
+        if(isNotBlank(characteristicValue)) {
```
alfonsomunozpomer: It’s better if you move the builder’s operations that are common to both the `if` and `else` clause above, your code will be easier to follow (also, test if you can have `.setNormalize(false)` before the conditional as well):
```java
var queryBuilder =
        new SolrQueryBuilder<SingleCellAnalyticsCollectionProxy>()
                .addQueryFieldByTerm(CHARACTERISTIC_NAME, characteristicName)
                .setFieldList(EXPERIMENT_ACCESSION);

if (isNotBlank(characteristicValue)) {
    ...
}
```
```

haideriqbal: Actually organising code made `else` code block disappear. And also setting `normalize` before the conditional works.

haideriqbal:
> Instead of putting all classes under the `experiments` package, use the `hcalandingpage` package to have all the feature self-contained. This strategy is called [‘package by feature’](https://dzone.com/articles/package-by-feature-is-demanded) and for both `atlas-web-bulk` and `atlas-web-single-cell` we’re trying to follow it. Also instead of `ScExperiment...` which is a bit non-descript try to be more specific (e.g. `HcaExperiment...` or `HumanExperiment...`.

I have renamed the classes to `HcaHuman...` and moved them to Hca Package. I thought first about renaming them to `Human...` but then `HumanExperimentDao` wasn't making sense to me because we are not getting just human accessions in the dao

PR #91, opened 2020-03-18
Fix HCA metadata endpoint and cache for for hca metada solr query

haideriqbal:
As the result of one part of solr stream query was larger than the default rows we set for solr query so we were missing on some of the information. Also as this query is taking a while to return the response (on my local machine it is taking around a minute). So added a cache for this method as well.


haideriqbal:
> This is an interesting one!
> 
> In our original stream expression [we set `qt=/export`, which is suggested when working with large result sets](https://lucene.apache.org/solr/guide/7_1/stream-source-reference.html#search-parameters); but with the Solr framework we have we can’t do this unless we used the constructor `CloudSolrStream(StreamExpression expression, StreamFactory factory)` and have that parameter in `factory`. To do this without breaking the Solr mini-framework as it is would be a lot of work, unfortunately.

I guess we should keep it in the back of our minds for the future.

PR #92, opened 2020-03-19
Chore #171839333 – Optimise t-SNE plot metadata JSON endpoint

alfonsomunozpomer:
I first pinpointed that the bulk (~95%) of the time spent by the metadata JSON endpoint was in the method `CellMetadataDao::getMetadataValueForCellIds`. This method retrieves the metadata values of a given characteristic type, for an experiment and its cell IDs. The Solr query we were building had three fields: an experiment accession, a metadata name (the sample characteristic) and all the experiments cell IDs. It was this third term the one that was causing that, for example, in E-HCAD-4, querying for e.g. `developmental stage` took about 68 seconds. The resulting query had an OR clause of ~600,000 terms. Just searching by experiment accession reduces the time to about five seconds.

I don’t know why we decided that passing the cell IDs besides the experiment accession was necessary. Do you see any reason why that would be the case? I’d like to know if you think the proposed change is fine.

Besides this, some other minor issues:
- I moved the `experiment_accession` from the query to the filters, [so result sets in Solr can be cached and re-used more effectively](https://lucene.apache.org/solr/guide/7_1/common-query-parameters.html#fq-filter-query-parameter).
- I noticed that we weren’t rounding the metadata endpoint to two decimal places like we do in the expression and cluster plot endpoints. For E-HCAD-4 I actually measured it and the difference is the predicted ~5MB. I think we established the ballpark figure of 6MB in a stand-up last week.


src/test/java/uk/ac/ebi/atlas/experimentpage/metadata/CellMetadataDaoIT.java
```diff
@@ -1,6 +1,5 @@
 package uk.ac.ebi.atlas.experimentpage.metadata;
 
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.BeforeAll;
```
haideriqbal: This class has two methods `validCellIdHasMetadataValues` and `validCellIdsHaveMetadataValues`. Isn't these two methods overlapping and should be merged in one?

src/test/java/uk/ac/ebi/atlas/experimentpage/tsneplot/TSnePlotServiceTest.java
```diff
@@ -92,9 +93,18 @@ void testFetchPlotWithMetadata() {
...
+                RandomDataTestUtils.generateRandomTSnePointDtos(NUMBER_OF_CELLS)
+                    .stream()
+                    .map(pointDto ->
+                            TSnePoint.Dto.create(
+                                    MathUtils.round(pointDto.x(), 2),
+                                    MathUtils.round(pointDto.y(), 2),
+                                    pointDto.name()))
+                    .collect(toImmutableSet());
+
+        when(tSnePlotDaoMock.fetchTSnePlotForPerplexity(eq(experimentAccession), eq(perplexity)))
```
haideriqbal: why do we need to use Mockito argument matcher `eq` here? Shouldn't we keep it simple and just pass `experimentAccession` and `perplexity`?

src/test/java/uk/ac/ebi/atlas/experimentpage/metadata/CellMetadataDaoIT.java
alfonsomunozpomer: They are testing different methods, but the names carry semantics of the old method, when we were passing multiple cell IDs. I’ve changed the name accordingly.

src/test/java/uk/ac/ebi/atlas/experimentpage/tsneplot/TSnePlotServiceTest.java
alfonsomunozpomer: Good catch. Maybe we were playing with arguments and left `eq` [because](https://site.mockito.org/javadoc/current/org/mockito/Mockito.html#3):
```java
verify(mock).someMethod(anyInt(), anyString(), eq("third argument"));
//above is correct - eq() is also an argument matcher

verify(mock).someMethod(anyInt(), anyString(), "third argument");
//above is incorrect - exception will be thrown because third argument is given without an argument matcher.
```

So if at any point we were doing a fuzzy match we set `eq` in the other argument and so on. That’s the only reason I can think of. In any case, simple is good. :smile:

PR #97, opened 2020-04-02
Move resource prop up from experimentTable

haideriqbal:
Now passing `resource` prop from `jsp` rather than in `FetchLoader`.


alfonsomunozpomer (approved):
All good here. Merge away. At some point I’d like to schedule a VC to show you how you prepare bundles for any given version of (Single Cell) Expression Atlas. What we discussed about your bundles not getting properly produced reminded me that this isn’t a simple process. The monorepo will eventually be a big help, though.

PR #99, opened 2020-05-14
Bugfix #172654192 – Values of metadata t-SNE plot are truncated

alfonsomunozpomer:
We were mis-labelling cells as _Not available_ because the Solr query had a row limit and the resulting map was missing many cell IDs.

I tried first to use the JSON facet API using two facets, one for `facet_characteristic_name` and another for `facet_factor_name`, since the resulting data structure bucketing by metadata value seemed more natural. However the time penalty was higher than I expected and took well over 30 seconds in big experiments, making that approach not good. You can try it yourself:
```bash
curl http://wp-np2-85:8983/solr/scxa-analytics/query -d '
rows=0&
q=*:*&
fl=cell_id,factor_value,characteristic_value&
json.facet={
  cells: {
    refine: true,
    type: terms,
    field: cell_id,
    limit: -1,
    domain: { filter: "experiment_accession:E-HCAD-4" }
    facet: {
      characteristic: {
        mincount: 1,
        refine: true,
        type: terms,
        field: facet_characteristic_value,
        limit: -1,
        domain: { filter: "characteristic_name:inferred_cell_type" }
      },
      factor: {
        mincount: 1,
        refine: true,
        type: terms,
        field: facet_factor_value,
        limit: -1,
        domain: { filter: "factor_name:inferred_cell_type" }
      }
    }
  }
}
'
```
That’s the reason why there are substantial modifications to the `SolrJsonFacetBuilder`. In any case the resulting change is an improvement, so I left it there. 

Then I tried the streaming expression approach and found a way to use [the `/export` handler](https://lucene.apache.org/solr/guide/7_1/stream-source-reference.html#search-parameters), which is ideal for our use case. :smile:

### DIFF
--- a/src/main/java/uk/ac/ebi/atlas/experimentpage/metadata/CellMetadataDao.java
+++ b/src/main/java/uk/ac/ebi/atlas/experimentpage/metadata/CellMetadataDao.java
@@ -132,6 +132,7 @@ public class CellMetadataDao {
 
         var queryBuilder =
                 new SolrQueryBuilder<SingleCellAnalyticsCollectionProxy>()
+                        .setRows(1500000)
                         .addFilterFieldByTerm(EXPERIMENT_ACCESSION, experimentAccession)
                         .addQueryFieldByTerm(fields)
                         .setFieldList(ImmutableSet.of(CELL_ID, CHARACTERISTIC_VALUE, FACTOR_VALUE));
--- a/src/main/java/uk/ac/ebi/atlas/experimentpage/metadata/CellMetadataDao.java
+++ b/src/main/java/uk/ac/ebi/atlas/experimentpage/metadata/CellMetadataDao.java
@@ -2,14 +2,16 @@ package uk.ac.ebi.atlas.experimentpage.metadata;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import org.apache.solr.common.SolrDocument;
 import org.springframework.stereotype.Component;
 import uk.ac.ebi.atlas.experimentimport.idf.IdfParser;
 import uk.ac.ebi.atlas.solr.cloud.SolrCloudCollectionProxyFactory;
+import uk.ac.ebi.atlas.solr.cloud.TupleStreamer;
 import uk.ac.ebi.atlas.solr.cloud.collections.SingleCellAnalyticsCollectionProxy;
-import uk.ac.ebi.atlas.solr.cloud.collections.SingleCellAnalyticsCollectionProxy.SingleCellAnalyticsSchemaField;
 import uk.ac.ebi.atlas.solr.cloud.search.SolrJsonFacetBuilder;
 import uk.ac.ebi.atlas.solr.cloud.search.SolrQueryBuilder;
+import uk.ac.ebi.atlas.solr.cloud.search.streamingexpressions.decorator.SelectStreamBuilder;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/ac/ebi/atlas/experimentpage/metadata/CellMetadataDao.java
```diff
@@ -19,21 +21,24 @@
...
+import static uk.ac.ebi.atlas.solr.cloud.collections.SingleCellAnalyticsCollectionProxy.FACET_FACTOR_VALUE;
 import static uk.ac.ebi.atlas.solr.cloud.collections.SingleCellAnalyticsCollectionProxy.FACTOR_NAME;
 import static uk.ac.ebi.atlas.solr.cloud.collections.SingleCellAnalyticsCollectionProxy.FACTOR_VALUE;
 import static uk.ac.ebi.atlas.solr.cloud.search.SolrQueryResponseUtils.extractSimpleOrderedMaps;
 
 @Component
 public class CellMetadataDao {
-    private SingleCellAnalyticsCollectionProxy singleCellAnalyticsCollectionProxy;
-    private IdfParser idfParser;
+    private final SingleCellAnalyticsCollectionProxy singleCellAnalyticsCollectionProxy;
```
haideriqbal, line 40: 👍

src/main/java/uk/ac/ebi/atlas/search/GeneSearchDao.java
```diff
@@ -143,27 +142,23 @@ public GeneSearchDao(NamedParameterJdbcTemplate namedParameterJdbcTemplate,
...
-                        .setFacetField(CHARACTERISTIC_NAME)
-                        .addSubFacets(ImmutableList.of(characteristicValueFacet))
-                        .setNestedFacet(true);
-
-        var facets =
+        var facetBuilder =
                 new SolrJsonFacetBuilder<SingleCellAnalyticsCollectionProxy>()
                         .setFacetField(EXPERIMENT_ACCESSION)
-                        .addSubFacets(ImmutableList.of(characteristicNameFacet));
+                        .addNestedFacet(
```
haideriqbal, line 148: 🎖️

src/main/java/uk/ac/ebi/atlas/experimentpage/metadata/CellMetadataDao.java
```diff
@@ -123,39 +128,43 @@ public CellMetadataDao(SolrCloudCollectionProxyFactory solrCloudCollectionProxyF
...
     // cells in an experiment
-    public Map<String, String> getMetadataValues(String experimentAccession,
-                                                 String metadataType) {
-        // We need to do this because we don't know if the metadata type is a factor or a characteristic
-        var fields = ImmutableMap.<SingleCellAnalyticsSchemaField, Collection<String>>of(
-                CHARACTERISTIC_NAME, ImmutableSet.of(metadataType),
-                FACTOR_NAME, ImmutableSet.of(metadataType));
+    public ImmutableMap<String, String> getMetadataValues(String experimentAccession,
+                                                            String metadataType) {
+        var METADATA_VALUE_TARGET_FIELD_NAME = "metadata_value";
```
haideriqbal, line 133: it would be better to make it a `static string` like `INFERRED_CELL_TYPE_SOLR_VALUE` and `SINGLE_CELL_IDENTIFIER_SOLR_VALUE`

alfonsomunozpomer: :+1:
Changed.

PR #100, opened 2020-05-29
Implement Experiment Collections functionality

haideriqbal:
Made a new dao class which query DB for experiment collections and return a `list`. As mentioned in `web-core` PR because of a change in `ExperimentJsonSerializer` class I had to make code changes to different classes and their test classes respectively.

Also added SQL scripts for testing purposes reflecting new experiment tables.

Accompanying `web-core` PR
https://github.com/ebi-gene-expression-group/atlas-web-core/pull/52


haideriqbal:
> What’s the class `ScxaExperimentCollections` doing? What’s its purpose and functionality?

I was confused about it because it just calls a dao function so instead of creating this class I can directly use dao class and make it as an implementation of ExperimentCollectionRepository but then I wasn't really sure if that would be right design approach. What would you suggest? Should I make dao directly implementing ExperimentCollectionRepository?

alfonsomunozpomer:
Sometimes you add a wrapper class to conform to some interface or API. If this isn’t the case, why don’t you get rid off `ScxaExperimentCollections`?

haideriqbal:
> Sometimes you add a wrapper class to conform to some interface or API. If this isn’t the case, why don’t you get rid off `ScxaExperimentCollections`?

Makes sense! I have deleted `ScxaExperimentCollections` and implemented `ExperimentCollectionReposity` in dao class.

src/test/java/uk/ac/ebi/atlas/experiments/ExperimentCollectionDaoIT.java
```diff
@@ -0,0 +1,51 @@
...
+    ExperimentCollectionDao subject;
+
+    @BeforeEach
+    void setUp() {
+        subject = new ExperimentCollectionDao(jdbcTemplate);
+    }
+
+    @Test
+    void ifNoExperimentCollectionReturnNothing() {
+        assertThat(subject.getExperimentCollections("E-FOO-0000"))
```
alfonsomunozpomer, line 40: You could very well use `getRandomExperimentAccession` and throw dice. :smile:

src/test/java/uk/ac/ebi/atlas/experiments/ExperimentCollectionDaoIT.java
```diff
@@ -0,0 +1,51 @@
...
+                .isEmpty();
+    }
+
+    @Test
+    void ifExperimentCollectionExistsReturnCollectionName() {
+        assertThat(subject.getExperimentCollections("E-EHCA-2"))
```
alfonsomunozpomer, line 46: If all our experiments are mapped to a collection I’d have an assertion in this test that picks one at random and check that the value is not empty. Just to have something a bit more general.

haideriqbal: I get your point. I am not changing the here at the moment because for now such test structure will pass in my local db but might fail on bamboo because atm not all experiments are mapped to a collection so if I pick up a one experiment accession at random then it might not have collection mapped to it resulting in failed test.

PR #105, opened 2020-07-13
Add a DAO returns cell types given an experiment..

lingyun1010:
accession and organism part


src/main/java/uk/ac/ebi/atlas/search/metadata/MetadataSearchDao.java
```diff
@@ -0,0 +1,92 @@
...
+import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.toMap;
+import static uk.ac.ebi.atlas.solr.cloud.collections.SingleCellAnalyticsCollectionProxy.CELL_ID;
+import static uk.ac.ebi.atlas.solr.cloud.collections.SingleCellAnalyticsCollectionProxy.CHARACTERISTIC_NAME;
+import static uk.ac.ebi.atlas.solr.cloud.collections.SingleCellAnalyticsCollectionProxy.CHARACTERISTIC_VALUE;
+import static uk.ac.ebi.atlas.solr.cloud.collections.SingleCellAnalyticsCollectionProxy.EXPERIMENT_ACCESSION;
+
+@Component
+public class MetadataSearchDao {
+    private static final Logger LOGGER = LoggerFactory.getLogger(FileDownloadController.class);
```
haideriqbal, line 24: Please change this to `MetadataSearchDao.class` in `getLogger`

src/main/java/uk/ac/ebi/atlas/search/metadata/MetadataSearchDao.java
```diff
@@ -0,0 +1,92 @@
...
+    private SingleCellAnalyticsCollectionProxy singleCellAnalyticsCollectionProxy;
+
+    public MetadataSearchDao(SolrCloudCollectionProxyFactory solrCloudCollectionProxyFactory) {
+        this.singleCellAnalyticsCollectionProxy =
+                solrCloudCollectionProxyFactory.create(SingleCellAnalyticsCollectionProxy.class);
+    }
+
+
+    private String getInferredCellTypeForCellId(String experimentAccession, String cellId) {
+        var characteristicFields = "inferred_cell_type";
```
haideriqbal, line 34: I believe this should be renamed to `characteristicField` as it is just single field name.

src/main/java/uk/ac/ebi/atlas/search/metadata/MetadataSearchDao.java
```diff
@@ -0,0 +1,92 @@
...
+                    .collect(Collectors.toList()).get(0);
+        } catch (Exception e) {
+            LOGGER.debug("Invalid cell id: {}", cellId);
+        }
+        return inferredCellType;
+    }
+
+    // Given a set of characteristic name (organism part) and characteristic value (pancreas),
+    // this method retrieves the value of that metadata for list of cell IDs.
+    @Cacheable(cacheNames = "cellIdsMetadata", key = "{#characteristicName, #characteristicValue}")
```
haideriqbal, line 56: There is no cache added in `CacheConfig.java`. Please add it there so that it can be registered with `cacheManager`

src/main/java/uk/ac/ebi/atlas/search/metadata/MetadataSearchDao.java
```diff
@@ -0,0 +1,92 @@
...
+public class MetadataSearchDao {
+    private static final Logger LOGGER = LoggerFactory.getLogger(FileDownloadController.class);
+    private SingleCellAnalyticsCollectionProxy singleCellAnalyticsCollectionProxy;
+
+    public MetadataSearchDao(SolrCloudCollectionProxyFactory solrCloudCollectionProxyFactory) {
+        this.singleCellAnalyticsCollectionProxy =
+                solrCloudCollectionProxyFactory.create(SingleCellAnalyticsCollectionProxy.class);
+    }
+
+
```
haideriqbal, line 32: Please remove an extra space.

src/test/java/uk/ac/ebi/atlas/search/metadata/MetadataSearchDaoIT.java
```diff
@@ -0,0 +1,75 @@
...
+@ContextConfiguration(classes = TestConfig.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class MetadataSearchDaoIT {
+    @Inject
+    private DataSource dataSource;
+
+    @Inject
+    private SolrCloudCollectionProxyFactory collectionProxyFactory;
+
+    private SingleCellAnalyticsCollectionProxy singleCellAnalyticsCollectionProxy;
```
haideriqbal, line 35: As this variable is just being used in `setUp` method so it would be better in my opinion to move it there as a local variable rather than a global variable.

src/main/java/uk/ac/ebi/atlas/search/metadata/MetadataSearchDao.java
upendrakumbham: As we are dealing with backend(Kind of data repository) and we mentioned DAO as part of the class name. So I would feel that @Repository  annotation suits more here than just independent @Component annotation.

upendrakumbham: In the search package, I could see 'GeneSearchDao' class. It is clearly informing the developer that this class deals with only genes. If we have any intention of that this class deals with only Cell metadata. We can rename the class to 'CellMetadataDao'

src/main/java/uk/ac/ebi/atlas/search/metadata/MetadataSearchDao.java
```diff
@@ -0,0 +1,92 @@
...
+                        toMap(
+                                Map.Entry::getKey,
+                                entry -> entry.getValue()
+                                        .stream()
+                                        .collect(
+                                                toMap(
+                                                        cell -> cell.get(CELL_ID.name()).toString(),
+                                                        cell -> getInferredCellTypeForCellId(
+                                                                entry.getKey(),
+                                                                cell.get(CELL_ID.name()).toString()
```
upendrakumbham, line 84: As we are calling 'getInferredCellTypeForCellId' method for each cell id. Would it be possible to use Streaming Expressions instead of a normal Solr query?

src/main/java/uk/ac/ebi/atlas/search/metadata/MetadataSearchDao.java
```diff
@@ -0,0 +1,92 @@
...
+        var results = this.singleCellAnalyticsCollectionProxy.query(queryBuilder).getResults();
+
+        var inferredCellType = "";
+        try {
+            inferredCellType = results
+                    .stream()
+                    .map(entry -> entry.get(CHARACTERISTIC_VALUE.name()).toString())
+                    .collect(Collectors.toList()).get(0);
+        } catch (Exception e) {
+            LOGGER.debug("Invalid cell id: {}", cellId);
```
upendrakumbham, line 49: Log an actual exception to help developer debugging using ERROR.

src/main/java/uk/ac/ebi/atlas/search/metadata/MetadataSearchDao.java
alfonsomunozpomer: This variable is just declared and used once. I think inlinining `"inferred_cell_type"` is better (fewer indirections when you read the code).

Always be wary of one-use variables.

src/main/java/uk/ac/ebi/atlas/search/metadata/MetadataSearchDao.java
alfonsomunozpomer: Just to clear things out a bit, we’ve never annotated DAO classes with `@Repository` because we’re not (or maybe it’s just me?) big fans of DDD. More seriously, since we’re not following any DDD principles because we’re old-school I can’t see the benefit of bothering with `@Component` subtypes (e.g. `@Service`), and they receive no special treatment. @upendrakumbham Is there any benefit of using `@Repository` (we’re no big Spring specialists)?

upendrakumbham: Since ages, I have got a habit of using @Component annotation for -  any Spring-managed component which doesn't belong to presentation, service, and data access layers. Somewhere I read about this in the spring documentation when I was learning spring. They would recommend using the more suitable stereotype annotations to improve readability as well as some special benefits in future releases for those annotations.
Yes, @Repository annotation has some special benefits to use.
[Stackoverflow Details](https://stackoverflow.com/questions/6827752/whats-the-difference-between-component-repository-service-annotations-in)
 @Component and @Service - currently both are same -  component scan and register a bean with the application context.
But we don't know, spring team may add some weightage to the @Service as well in the future releases :).

alfonsomunozpomer: I see! Thanks for the links. We can add `@Repository` to all of our DAO classes (as we go about adding and updating them, no rush to do it), with the understanding that in our project DAO and DDD repositories are pretty much the same.

alfonsomunozpomer:
I merged `master` since the branch where this PR was based was a bit outdated. I’d suggest you see the changes in each commit after the merge. Otherwise you’re going to see a few changes unrelated to this PR.

As for actual changes, instead of the general metadata search DAO previously implemented, I implemented a Solr query that pulls all cell types given an ontology URI. This is the query we need for the cell type heatmap in the experiment page. The query uses the implementation for `CartesianProductStream` we discussed, in https://github.com/ebi-gene-expression-group/atlas-web-core/pull/55, so please review that PR first if you haven’t.

For this new query I needed to update the SDRF files in our test datasets, which forced me to re-download them and re-import them into Solr, and then I needed to use the new `inferred_cell_type_-_ontology_labels` metadata value instead of the old `inferred_cell_type`. Which meant I also needed to make adjustments in other classes and tests accordingly. [I talked about this with Anja on Slack](https://ebi-fg.slack.com/archives/C800ZEPPS/p1601122066011300), since certain assumptions we made in our code were incorrect. The last four commits are related to this.

One thing I’d like is to have better tests for the new class. If you could add something else it would be helpful.

src/main/java/uk/ac/ebi/atlas/experimentpage/metadata/CellMetadataDao.java
```diff
@@ -90,10 +90,10 @@ public CellMetadataDao(SolrCloudCollectionProxyFactory solrCloudCollectionProxyF
         return characteristics.build();
     }
 
-    public Map<String, String> getMetadataValuesForCellId(String experimentAccession,
-                                                          String cellId,
-                                                          Collection<String> factorFields,
-                                                          Collection<String> characteristicFields) {
+    public ImmutableMap<String, String> getMetadataValuesForCellId(String experimentAccession,
```
haideriqbal, line 93: 👍

src/test/java/uk/ac/ebi/atlas/search/CellTypeSearchByOrganismPartDaoIT.java
```diff
@@ -0,0 +1,74 @@
...
+                .isNotEmpty();
+        // islet of Langerhans
+        var cellTypesInIsletOfLangerhans = subject.getCellTypeMetadata("E-MTAB-5061", "http://purl.obolibrary.org/obo/UBERON_0000006");
+        assertThat(cellTypesInIsletOfLangerhans)
+                .isNotEmpty();
+
+        assertThat(cellTypesInPancreas)
+                .containsAll(cellTypesInIsletOfLangerhans);
+    }
+}
```
haideriqbal, line 74: I will also add one more test case in the scenario where we search for two `organsOrOrganismPart` which are not related to each other and should not contain each other `cell types`. For example, searching for `skin` and `kidney` should not have common cell types. I hope it makes sense.

src/main/java/uk/ac/ebi/atlas/configuration/CacheConfig.java
```diff
@@ -34,6 +34,7 @@ public CacheManager cacheManager() {
                 builder -> builder.name("jsonTSnePlotWithClusters"),
                 builder -> builder.name("jsonTSnePlotWithMetadata"),
 
-                builder -> builder.name("hcaMetadata"));
+                builder -> builder.name("hcaMetadata"),
+                builder -> builder.name("cellIdsMetadata"));
```
haideriqbal, line 38: It should be `cellTypes` instead of `cellIdsMetadata`.

haideriqbal (changes requested):
Just a couple of comments. Please review!

src/test/java/uk/ac/ebi/atlas/search/CellTypeSearchByOrganismPartDaoIT.java
alfonsomunozpomer: Have we got any experiment with multiple organs in our test datasets? I don’t feel like loading Tabula Muris in our laptops. :D

haideriqbal: 1_ZyE6yAwn5DlOfsO6UTC8Yw

upendrakumbham (approved):
The approach and implementation details looks clean for me. I can see that you have already fixed @haideriqbal comment about cache.

src/test/java/uk/ac/ebi/atlas/search/CellTypeSearchByOrganismPartDaoIT.java
alfonsomunozpomer: So I played a bit with the test experiment @upendrakumbham mentioned, E-EHCA-2, and it turns out that the test case mentioned above doesn’t hold.

For example, the following query shows the `inferred cell type - authors labels` from a set of skin cells:
```solr
cell_id:(21784_6_131 OR 21784_6_137 OR 21784_6_169 OR 21784_6_181 OR 21784_6_183 OR 21784_6_191 OR 21784_6_225 OR 21784_6_227 OR 21784_6_231 OR 21784_6_267 OR 21784_6_279 OR 21784_6_321 OR 21784_6_329 OR 21784_6_33 OR 22467_8_315 OR 22467_8_321 OR 22467_8_322 OR 22467_8_324 OR 22467_8_332 OR 22467_8_343 OR 22467_8_345 OR 22467_8_348 OR 22467_8_349 OR 22467_8_355 OR 22467_8_356 OR 22467_8_368 OR 22467_8_375 OR 22467_7_94 OR 22467_8_194 OR 22467_8_195 OR 22467_8_197 OR 22467_8_200 OR 22467_8_202 OR 22467_8_206 OR 22467_8_208 OR 22467_8_211 OR 22467_8_218 OR 22467_8_222 OR 22467_8_224 OR 22467_8_228 OR 22467_8_241 OR 22467_8_243 OR 22467_8_246 OR 22467_7_44 OR 22467_7_45 OR 22467_7_46 OR 22467_7_47 OR 22467_7_52 OR 22467_7_60 OR 22467_7_64 OR 22467_7_67 OR 22467_7_68 OR 22467_7_69 OR 22467_7_75 OR 22467_7_77 OR 22467_7_78 OR 22467_7_79 OR 22467_7_82 OR 22467_7_86 OR 22467_7_87 OR 22467_7_89 OR 22467_7_91 OR 22467_7_317 OR 22467_7_318 OR 22467_7_319 OR 22467_7_320 OR 22467_7_331 OR 22467_7_336 OR 22467_7_337 OR 22467_7_343 OR 22467_7_35 OR 22467_7_358 OR 22467_7_36 OR 22467_7_360 OR 22467_7_361 OR 22467_7_37 OR 22467_7_373 OR 22467_7_374 OR 22467_7_376 OR 22467_8_252 OR 22467_8_253 OR 22467_8_254 OR 22467_8_257 OR 22467_8_258 OR 22467_8_259 OR 22467_8_262 OR 22467_8_270 OR 22467_8_271 OR 22467_8_288 OR 22467_8_292 OR 22467_8_294 OR 22467_8_306 OR 22467_8_308 OR 22467_7_214 OR 22467_7_216 OR 22467_7_218 OR 22467_7_220 OR 22467_7_222 OR 22467_7_225 OR 22467_7_226) AND facet_characteristic_name:(inferred_cell_type_-_authors_labels OR organism_part)
```

Among the results we have: `B cell` or `lymph node T cell`.

The same query for a set of lymph node cells:
```solr
cell_id:(21935_5_245 OR 21935_5_246 OR 21935_5_247 OR 21935_5_248 OR 21935_5_250 OR 21935_5_251 OR 21935_5_253 OR 21935_5_258 OR 21935_5_266 OR 21935_5_274 OR 21935_5_280 OR 21935_5_293 OR 21935_5_3 OR 21935_5_300 OR 21935_5_302 OR 21935_5_306 OR 21935_5_307 OR 21935_5_309 OR 21935_5_316 OR 21935_5_32 OR 21935_5_322 OR 21935_5_328 OR 21935_5_345 OR 21935_5_348 OR 21935_5_352 OR 21935_5_36 OR 21935_5_366 OR 21935_5_42 OR 21935_5_44 OR 21935_5_48 OR 21935_5_49 OR 21935_5_55 OR 21935_5_60 OR 21935_5_61 OR 21935_5_62 OR 21935_5_68 OR 21935_5_72 OR 21935_5_8 OR 21935_5_82 OR 21935_5_86 OR 21935_5_90 OR 21935_5_98 OR 21935_6_103 OR 21935_6_105 OR 21935_6_12 OR 21935_6_121 OR 21935_6_123 OR 21935_6_125 OR 21935_6_126 OR 21935_6_127 OR 21935_6_149 OR 21935_6_150 OR 21935_6_151 OR 21935_6_154 OR 21935_6_159 OR 21935_6_161 OR 21935_6_162 OR 21935_6_163 OR 21935_6_17 OR 21935_6_175 OR 21935_6_177 OR 21935_6_179 OR 21935_6_18 OR 21935_6_182 OR 21935_6_193 OR 21935_6_20 OR 21935_6_203 OR 21935_6_204 OR 21935_6_205 OR 21935_6_222 OR 21935_6_228 OR 21935_6_235 OR 21935_6_236 OR 21935_6_238 OR 21935_6_239 OR 21935_6_240 OR 21935_6_245 OR 21935_6_249 OR 21935_7_332 OR 21935_7_34 OR 21935_7_342 OR 21935_7_350 OR 21935_7_40 OR 21935_7_42 OR 21935_7_54 OR 21935_7_56 OR 21935_7_60 OR 21935_7_64 OR 21935_7_90 OR 21935_8_100 OR 21935_8_106 OR 22011_1_167 OR 22011_1_17 OR 22028_2_109 OR 22028_2_115 OR 22028_2_119 OR 22028_2_121 OR 22028_2_128 OR 22028_2_134 OR 22028_2_140) AND facet_characteristic_name:(inferred_cell_type_-_authors_labels OR organism_part)

**Review by upendrakumbham — CHANGES_REQUESTED** [review]
I think it would be nice if we add a test to prove the behavior of CellTypeMetadata() in case of both ontology labels and authors labels search returns empty cell types for the organism part.

File: src/main/java/uk/ac/ebi/atlas/configuration/CacheConfig.java [hunk 11]
```diff
@@ -35,6 +35,7 @@ public CacheManager cacheManager() {
                 builder -> builder.name("jsonTSnePlotWithMetadata"),
 
                 builder -> builder.name("hcaMetadata"),
-                builder -> builder.name("cellTypes"));
+                builder -> builder.name("inferredCellTypesOntology"),
+                builder -> builder.name("inferredCellTypesAuthors"));
```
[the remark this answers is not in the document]
alfonsomunozpomer: Yes, in principle I also thought of using the same cache, but if one experiment is missing one type of annotations, an empty set would be cached and at the service level we wouldn’t be able to implement the logic “check to see if there are ontology labels, if not use authors labels instead”.

src/test/java/uk/ac/ebi/atlas/search/CellTypeSearchByOrganismPartDaoIT.java
```diff
@@ -24,51 +19,57 @@
...
+                        "E-MTAB-5061", "http://purl.obolibrary.org/obo/UBERON_0001264");
+        var authorsLabelsInPancreas =
+                subject.getInferredCellTypeAuthorsLabels(
+                        "E-MTAB-5061", "http://purl.obolibrary.org/obo/UBERON_0001264");
+
+        assertThat(ontologyLabelsInPancreas)
+                .doesNotContainSequence(authorsLabelsInPancreas);
+        assertThat(ontologyLabelsInPancreas)
+                .containsAnyElementsOf(authorsLabelsInPancreas);
+    }
```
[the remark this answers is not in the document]
alfonsomunozpomer: What would such test accomplish? Other than such datasets can or cannot exist (which is a test that should be part of the curation and data production layer). Also, I’m not sure we’ve got datasets without those annotations. Could you ask curators to know if that’s the case? At least so we know what to expect when we add the service on top.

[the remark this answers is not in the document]
upendrakumbham: Hi @alfonsomunozpomer, Sorry if I confuse you. We should have both ontologies and authors' labels in the dataset. 
My point is that as you implemented a fallback method. Which method is the highest priority to call from the DB DAO or Service layer. If that gets empty cell types then call fallback one from the service or DB DAO class.
 I have mentioned that to add a negative test case to know these priority levels. Pick an organism part to get empty cell types for the "getInferredCellTypeOntologyLabels". Then use fall back method to fetch cell types..etc. Does it makes sense :)?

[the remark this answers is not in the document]
alfonsomunozpomer: Yes, the fallback method is to be used it if the ontology labels are empty, but that’s something that the service layer should do. The DAO shouldn’t be concerned with business logic. Let the service choose what to do if no ontology labels are available.

alfonsomunozpomer (approved):
I did this so I’m happy with it. :D

upendrakumbham (approved):
Everything looks good to me 👍.

PR #106, opened 2020-08-09
Hotfix #173033902 – Improve gene search performance

alfonsomunozpomer:
It turns out that having a clause in Postgres of the form `SELECT ... WHERE foo IN (SELECT ...))` incurs in a big performance hit. Even if both the main query and the nested one separately complete in the order of 0.01 seconds, combining the two results in times ranging from 0.5 to 3 seconds. This was the case when looking for the marker gene profile of a gene ID (i.e. for which k and cluster ID in an experiment the gene ID is a marker gene, or none).

In the case of popular genes that appear in many experiments, we were looking for these marker gene profiles without caching and doing a `SELECT` query for each experiment, resulting in gene searches that made more than 40 queries (of 0.5-3 seconds each). Also, as part of this step, the IDF file was parsed each time to find the preferred K of the experiment. 

To solve the above:
• I added caching to avoid parsing the IDF each time
• Instead of finding the minimum marker probability of each gene ID in an experiment every time, I added a method that creates (and caches) a map of gene ID to minimum marker probability

I thought of adding the cache in the service level, but that would mean that other genes in the same experiment would suffer long first access times, and gene searches are very variable.

When this PR is accepted we must remember to cherry-pick it in `master` as well.

There is a detailed report with times and benchmarks at https://www.pivotaltracker.com/story/show/173033902.


upendrakumbham:
I have gone through the code to know what magic you did to boost SQL query performance. Now I got that magic :).
Everything looks good to me.

PR #109, opened 2020-09-14
Bugfix #174786171 – Supplementary Information tab is broken

alfonsomunozpomer:
There was an error in rendering because requests to URLs such as `http://localhost:8080/gxa/sc/json/experiments/E-GEOD-99058/resources/SUPPLEMENTARY_INFORMATION` returned a 500 error.

The request mapping method is returning a String which is interpreted as a Spring MVC view becuase the `@ResponseBody` annotation is missing (compare with bulk Expression Atlas). I replaced `@Controller` with `@RestController` to implicitly add `@ResponseBody`.

### DIFF
--- a/src/main/java/uk/ac/ebi/atlas/experimentpage/ExternallyAvailableContentController.java
+++ b/src/main/java/uk/ac/ebi/atlas/experimentpage/ExternallyAvailableContentController.java
@@ -3,10 +3,10 @@ package uk.ac.ebi.atlas.experimentpage;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonObject;
 import org.springframework.http.MediaType;
-import org.springframework.stereotype.Controller;
 import org.springframework.web.bind.annotation.GetMapping;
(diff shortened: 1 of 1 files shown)


lingyun1010:
I am a little bit confused as I tested on [dev](https://wwwdev.ebi.ac.uk/gxa/sc/json/experiments/E-GEOD-99058/resources/SUPPLEMENTARY_INFORMATION) and it works. If the error is only on local server?

alfonsomunozpomer:
Yes, I’m not really sure what is deployed, but it fails only in `localhost`. Can you please double check the code in `master` to make sure I’m missing something?

lingyun1010:
I have checked the `202006` deployment branch, and it uses `@ResponseBody`. I am afraid that it was my fault, while I was debugging the new release and committed directly to that branch.

lingyun1010 (approved):
Thanks for this debug and I will be careful in the next release. LGTM.

PR #120, opened 2021-01-19
Features #174379278, #174379548, #173919292 – Integrate general heatmap and anatomogram in experiment results tabs

alfonsomunozpomer:
I added two methods in the `HeatmapHighchartsAdapter` to sort labels lexicographically or numerically depending on the typee of heatmap we want: the former for cell types and the latter for clusters. Otherwise clusters would be shown in the order 1, 10, 11, 12, ..., 19, 20, 21, 22, ..., 29, 3, 4, 5...

I imported the beta release of the general heatmap and used both in the anataomogram tab and in the marker genes heatmap.


src/test/java/uk/ac/ebi/atlas/experimentpage/JsonAnatomogramCellTypeHeatmapControllerWIT.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class JsonAnatomogramCellTypeHeatmapControllerWIT {
+
+    @Test
+    void foo() {
+        assertThat(true).isFalse();
```
upendrakumbham: I was looking for a test case to test that relevant endpoint. I am confused here, what we are testing here @alfonsomunozpomer.

upendrakumbham (approved):
All LGTM! minor thing, a clarification on 

> 'JsonAnatomogramCellTypeHeatmapControllerWIT'

 test case.

src/test/java/uk/ac/ebi/atlas/experimentpage/JsonAnatomogramCellTypeHeatmapControllerWIT.java
alfonsomunozpomer: Oh, I configured IDEA with that template, so that whenever I create new classes with their corresponding tests, by default it will always fail. In that way, if I forget to add a test it will always fail. In this case this was from a class `JsonAnatomogramCellTypeHeatmapController` that ended being not relevant and I removed the source class but the test remained.

src/test/java/uk/ac/ebi/atlas/search/OntologyAccessionsSearchServiceIT.java
```diff
@@ -0,0 +1,36 @@
...
+
+    private OntologyAccessionsSearchService subject;
+
+    @BeforeEach
+    void setUp() {
+        subject = new OntologyAccessionsSearchService(ontologyAccessionsSearchDao);
+    }
+
+    @Test
+    void foo() {
```
haideriqbal, line 32: Please change the test case name.

src/test/java/uk/ac/ebi/atlas/search/OntologyAccessionsSearchDaoIT.java
```diff
@@ -0,0 +1,37 @@
...
+    private OntologyAccessionsSearchDao subject;
+
+    @BeforeEach
+    void setUp() {
+        subject = new OntologyAccessionsSearchDao(collectionProxyFactory);
+    }
+
+
+    @Test
+    void foo() {
```
haideriqbal, line 34: Here as well!

src/main/java/uk/ac/ebi/atlas/experimentpage/markergenes/MarkerGenesDao.java
```diff
@@ -11,9 +10,9 @@
 @Repository
 public class MarkerGenesDao {
     private final NamedParameterJdbcTemplate namedParameterJdbcTemplate;
-    private static final String CELL_GROUP_TYPE = "inferred cell type";
+	private static final String CELL_GROUP_TYPE = "inferred cell type - ontology labels";
```
haideriqbal, line 13: Shouldn't this be `inferred cell type _-_ ontology labels`? As this is how they are stored in Solr

haideriqbal (approved):
Few minor comments. Otherwise LGTM!

lingyun1010 (approved):
LGTM

src/main/java/uk/ac/ebi/atlas/experimentpage/markergenes/MarkerGenesDao.java
alfonsomunozpomer: We are querying the database, where they are stored with spaces. Confusing enough... It’d be good to try to normalise this though. Can you add a story to the backlog, please? This needs to be discussed with data production, too.

haideriqbal: I will add the story!

PR #123, opened 2021-01-20
Optimise /json/experiments endpoint

haideriqbal:
Mitch reported that /json/experiments endpoint was taking ~4 secs to return the experiment list. Upon investigation of the code, it was found that the results were not being cached. I added a new cache and also evicted it at the proper places in the experiment CRUD to keep the results consistent.


alfonsomunozpomer (approved):
All good. Merge away!

Issue #127, opened 2021-01-21
Do not show anatomogram for experiments with no displayable data


sfexova:
At the moment anatomograms are displayed whenever the [organism] is human and the [organism part] is one for which an anatomogram exists even if there is no data that can actually be displayed in the relevant anatomogram:
e.g
https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-GEOD-130473/downloads - is fetal liver (have anatomograms for adult tissues only)
https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-HCAD-8/results/anatomogram - T-cells not part of the diagrams
https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-MTAB-6653/results/anatomogram - don't have inferred cell types
https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-GEOD-86618/results/anatomogram - don't have inferred cell types
https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-CURD-11/results/anatomogram - don't have inferred cell types
https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-CURD-11/results/anatomogram - don't have inferred cell type
https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-HCAD-10/results/anatomogram - only have fetal data (not adult)
https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-HCAD-10/results/anatomogram - don't have inferred cell types

The best thing would be removing the anatomograms from these experiments before the public release if at all possible as displaying the anatomograms for these is not correct but a lasting change to the implementation logic may be too complicated to do in that time-frame - if yes, can we 'manually' remove them for now, before a permanent solution is found?

For future development I thing the logic for display should be [organism] 'adult', [organism part] - anatomogram available, [inferred cell type - ontology labels] - available AND some terms therein or their parent terms are used in the relevant anatomogram stack
But not sure how difficult this would be to implement (especially the last bit as that assumes ontology expansion for each ontology term used in an anatomogram is somehow available to check against)
Also for placenta the [dev stage] is tricky as it's a part fetal/ part mother (adult) organ but that should be the only exception where the [dev stage] check would not be applicable

pcm32:
[...]

PR #148, opened 2021-02-03
Hide Anatomogram with incomplete data

haideriqbal:
For experiments with incomplete anatomogram data, they should not display anatomograms. See issue here:

https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/127

### DIFF
--- a/src/main/java/uk/ac/ebi/atlas/experimentpage/tabs/ExperimentPageContentService.java
+++ b/src/main/java/uk/ac/ebi/atlas/experimentpage/tabs/ExperimentPageContentService.java
@@ -2,6 +2,7 @@ package uk.ac.ebi.atlas.experimentpage.tabs;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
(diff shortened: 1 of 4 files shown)


lingyun1010 (changes requested):
I have tested it locally, and the `anatomogram` props is required in the front-end router, so I think you are right that we need to pass an empty value for `anatomogram` props. Otherwise, the result page will not load. Please take care of the test correspondingly, instead of missing `anatomogram` but has an empty value for those exception experiments.

lingyun1010 (approved):
LGTM

src/main/java/uk/ac/ebi/atlas/experimentpage/tabs/ExperimentPageContentService.java
```diff
@@ -70,11 +82,16 @@ public JsonObject getTsnePlotData(String experimentAccession) {
...
+        if (EXPERIMENTS_WITH_NO_ANATOMOGRAM.contains(experimentAccession)) {
+            result.add(
+                    "anatomogram", GSON.toJsonTree(""));
+        } else {
+            result.add(
+                    "anatomogram",
+                    GSON.toJsonTree(
+                            ontologyAccessionsSearchService.searchAvailableAnnotationsForOrganAnatomogram(
+                                    experimentAccession)));
+        }
```
alfonsomunozpomer, line 94: Remember that you can rewrite this as:
```java
result.add(
        "anatomogram",
        GSON.toJsonTree(EXPERIMENTS_WITH_NO_ANATOMOGRAM.contains(experimentAccession) ?
                "" :
                ontologyAccessionsSearchService.searchAvailableAnnotationsForOrganAnatomogram(experimentAccession));
```

Now, I’m not saying that you should change it, but consider it before merging and write it however it feels more expressive and easier to understand to you.

haideriqbal: Sounds good Alfonso. I've changed it according to the suggested style. It looks much cleaner and also avoids code duplication.

Issue #121, opened 2021-01-19
On landing on an anatomogram sub-level, all contained cells should be shown in the heatmap


pcm32:
When going into experiment https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-GEOD-130148/results

and clicking on 

![image](https://user-images.githubusercontent.com/368478/105088605-05914c00-5a94-11eb-93fb-dfde7dee1ceb.png)

you reach a page with no cell types in the heatmap.

![image](https://user-images.githubusercontent.com/368478/105088513-e4c8f680-5a93-11eb-8f04-3b206ef59702.png)

However, clicking individual cells

![image](https://user-images.githubusercontent.com/368478/105088704-29ed2880-5a94-11eb-9ac0-320f3cf88250.png)

will show that they do have gene expression data

![image](https://user-images.githubusercontent.com/368478/105088744-396c7180-5a94-11eb-9451-0727c7ef3fc3.png)

which probably means that the first previou marker genes heatmap could be populated with the marker genes of each cell type as rows and different cell types as columns.

pcm32:
For completeness, the rest of what I commented on slack:

When I click here (on the starting anatomogram):
![image](https://user-images.githubusercontent.com/368478/105149731-c696e100-5afb-11eb-8803-b377af53887b.png)

I get here, which looks as expected (has the filled heatmap with different cell types as columns):
![image](https://user-images.githubusercontent.com/368478/105149266-43758b00-5afb-11eb-9a06-8d02be34e9b1.png)

although then on the next rectangle click on that page, it gets us to again the initially empty marker gene heatmap case:

![image](https://user-images.githubusercontent.com/368478/105149391-66a03a80-5afb-11eb-9fe1-c9681437080a.png)

Is it that when we don’t have next depth figures you get the issue? So figures that don’t have blue rectangles to click on initially will display with the heatmap empty. There is maybe an issue on how the query is done for those I guess.

pcm32:
I'm adding this to the milestone under the assumption that is doable @alfonsomunozpomer , although I'm not sure if an explanation for this has been found yet. We can discuss when we have a better understanding.

PR #149, opened 2021-02-03
Feature #176761932 – Send multiple onotology uris to back end

alfonsomunozpomer:
I expanded the views which weren’t covered by a single ontology ID to the correspondin arrays and passed a set of values in the request parameter `organismPart`.


lingyun1010:
It looks good, but I did not find the way to check ontologies parent-children relationship on the OLS so I did not make any progress on the views. Do you need me to double check if the ontologies you expand is correct or not?

lingyun1010 (approved):
Please merge!

pcm32:
This deals with #121

PR #150, opened 2021-02-05
Bugfix #176761819, #176817876 – Reset heatmap whenever the anatomogram view is changed; back icon works without any selection

alfonsomunozpomer:
Includes https://github.com/ebi-gene-expression-group/organ-anatomogram/pull/3 and https://github.com/ebi-gene-expression-group/organ-anatomogram/pull/4.

Use `afterSwitchView` to request data for the cell type heatmap, taking into consideration if a single ontology URI can cover all the cell types in the specified view.


lingyun1010 (approved):
LGTM!

PR #161, opened 2021-03-29
Let HCA endpoint fail gracefully

haideriqbal:
We had an issue where when an experiment could not be found in the db the endpoint threw a `404` error which was not useful at all. So now if such a scenario occurs, the endpoint would throw a `503` error.

### DIFF
--- a/src/main/java/uk/ac/ebi/atlas/hcalandingpage/HcaHumanExperimentsController.java
+++ b/src/main/java/uk/ac/ebi/atlas/hcalandingpage/HcaHumanExperimentsController.java
@@ -1,11 +1,12 @@
 package uk.ac.ebi.atlas.hcalandingpage;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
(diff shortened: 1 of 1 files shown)


upendrakumbham (changes requested):
Hi @haideriqbal,
This is only my view with a REST API level of experience :). We can discuss further this approach if you like:

Returning the `503` HTTP status code confuses the user. Because GET `/json/experiments/hca/human` rest endpoint responsibility-`retrieves human experiments from the backend`. It's doing its work. We should return '200' status even though we have some data discrepancies.

Issue: We have some data discrepancies, users not really required to know this, we should handle this internally in our team. So we need to find a way to log these missing experiments:

For me, the good place is the @service layer

1. Compare Solr and database experiments
2. Find DB missing experiments
3. Log them to the developers. The developer's responsibility to highlight these missing experiments with the data production team to make data corrections further where ever required

haideriqbal:
Really good solution. If the error occurs now, it sends this response `{"error":"Experiment with accession E-CURD-10 could not be found"}` which is what we want. If you are happy @upendrakumbham please approve the PR.

upendrakumbham (approved):
LGTM!

PR #163, opened 2021-04-23
 Chore #177275244 – Migration to Gradle 7 with multi-project layout

alfonsomunozpomer:
Many big changes!

All the code has been moved to `app` and `atlas-web-core` has been added as a Git submodule. In this way we can create a Gradle multi-project build in which `atlas-web-core` is a non-versioned dependency.

A new Gradle plug-in that contains all the common build logic for all our `atlas-web-*` projects has been created. You can find it in `buildSrc`. I don’t find the directory structure very appealing, but [it’s dictated by Gradle](https://docs.gradle.org/current/userguide/declaring_dependencies_between_subprojects.html). I removed the `build-profiles` submodule, since now all build logic has been encapsulated in the plug-in and when testing, you can specify settings such as Solr host or DB credentials from Gradle itself. For instance, when testing in Bamboo the command is:
```bash
./gradlew -PsolrHost=lime ...
```

Whereas in your dev laptop it will be:
```bash
./gradlew -PsolrHost=localhost` ...
```

If nothing is provided there are some defaults that you can check in `buildSrc/src/main/groovy/atlas-web-app.java-conventions.gradle`. Notice that properties like `dataFilesLocation` and `experimentFilesLocation` rely on the environment `ATLAS_DATA_PATH` to be set. In this way we are consistent with the way the Docker Compose instance works.

Another signficant change is that properties files have been removed from `main/resources` and are only provided in `test/resources`. When deploying SCXA we’ll inject those files via Tomcat’s context file. There’s a new `context.xml` suitable for Docker, but it will be ignored in any of our hosted environments (test, staging, public and fallback) because there’s a `gxa#sc.xml` file which points at the properties files there that takes precedence. You’ll see a new `docker/atlas-properties` directory which is bind-mounted on the `scxa-tomcat` container.


app/build.gradle
```diff
@@ -0,0 +1,47 @@
+plugins {
+    id 'atlas-web-app.java-conventions'
```
lingyun1010, line 2: As far as I know, the build logic in `atlas-web-app.java-conventions` is not shared by any other projects in this repo. Is it necessary to have a seperate gradle build file here?

lingyun1010: I got the answer in our meeting. Since I have not pulled the latest submodule, I don't have this plugin in web-core project.

buildSrc/build.gradle
```diff
@@ -0,0 +1,4 @@
+plugins {
+    id 'groovy-gradle-plugin'
```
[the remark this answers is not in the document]
alfonsomunozpomer: This is necessary to indicate that `buildSrc` contains the definition of a Gradle plugin.

PR #165, opened 2021-05-26
178200485/switch to test containers

upendrakumbham:
This PR target: 

- Get rid of existing H2 database to run test cases. 

- Add TestContainers  and flyway libraries in the scxa `build.gradle` file.

- Use those features in the code .Test Containers creates Postgres container, creates empty schema using flyway library and runs all the tests and TestContainers takes care of destroying container and it's resources on it's own after tests finished.

`-Notice:`
Bamboo CI build is not passed for this branch, as this VM is very old , it's doesn't have docker supporting environment as per @alfonsomunozpomer, as we are moving to Jenkins soon, this reviewer has to test this PR in their local machine to make sure that all tests passing with TestContainers.


build.gradle
```diff
@@ -30,6 +30,7 @@ repositories {
 }
 
 def cache2kVersion = '1.2.4.Final'
+def junitJupiterVersion = '5.6.0'
```
alfonsomunozpomer, line 33: We can use a BOM instead:
```
testImplementation platform('org.junit:junit-bom:5.6.0')
```

src/test/java/uk/ac/ebi/atlas/configuration/AbstractContainerBaseTest.java
```diff
@@ -0,0 +1,31 @@
...
+
+import com.zaxxer.hikari.HikariConfig;
+import com.zaxxer.hikari.HikariDataSource;
+import org.flywaydb.core.Flyway;
+import org.testcontainers.containers.JdbcDatabaseContainer;
+import org.testcontainers.containers.PostgreSQLContainer;
+
+import javax.sql.DataSource;
+
+abstract class AbstractContainerBaseTest {
```
alfonsomunozpomer, line 11: Why is this class abstract?
Why does it have `Abstract` in the name when it’s abstract already?
Why do we need this class?

build.gradle
```diff
@@ -95,6 +96,11 @@ dependencies {
     implementation 'jstl:jstl:1.2'
     // http://commons.apache.org/proper/commons-configuration/dependencies.html
     implementation 'commons-beanutils:commons-beanutils:1.9.3'
+    implementation 'com.zaxxer:HikariCP:3.3.1'
+    // https://mvnrepository.com/artifact/org.flywaydb/flyway-core
+    implementation group: 'org.flywaydb', name: 'flyway-core', version: '7.2.0'
```
alfonsomunozpomer, line 101: These two dependencies could (and should) be `testImpementation`.

build.gradle
```diff
@@ -95,6 +96,11 @@ dependencies {
...
+    //This takes care of 'TestContainer' versioning
+    implementation platform('org.testcontainers:testcontainers-bom:1.15.3') //import bom
```
alfonsomunozpomer, line 103: Again, this would better be `testImplementation`.

Also, be careful about overcommenting. `//import bom` is precisely what the line is doing, it adds no information. I remember one textbook example which was something like this:
```c
return a + b;  // Add a and b and return the sum
```

src/test/java/uk/ac/ebi/atlas/configuration/AbstractContainerBaseTest.java
```diff
@@ -0,0 +1,31 @@
...
+    static final PostgreSQLContainer TEST_DATABASE;
+    static final String DOCKER_IMAGE_VERSION = "postgres:10-alpine";
+    static final String FLY_WAY_PATH = "filesystem:./src/test/resources/schemas/flyway/scxa";
```
alfonsomunozpomer, line 14: Notice that these strings are used only once. I think it’s better in that case to inline them, so there are no indirections when reading the code.

alfonsomunozpomer (approved):
Good job!

From these changes it’s been a breeze to create the PR in the Gitflow version of the project. Please have a look at https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/166. There are some rough edges that I commented here and that are fixed there.

One final comment, be aware about our naming convention for branches: `feature/<story no>-<branch-name>`. Be careful about this since we want to have an easy way to navigate the Git branches in the new project layout. Be sure to install the Gitflow extensions if that helps.

I’m really happy to have a real PostgreSQL DB in place now. This also paves the way towards having a Solr container and then all our tests can be self-contained.

src/test/java/uk/ac/ebi/atlas/configuration/AbstractContainerBaseTest.java
upendrakumbham: We don't need this class. I tried to implement it in the TestConfig class, but I got some issues. So ended up this class.

build.gradle
upendrakumbham: Correct, I'm using them only for tests.

build.gradle
upendrakumbham: Should be `testImplementation` Sorry, copy-paste error. Will take care.

src/test/java/uk/ac/ebi/atlas/configuration/AbstractContainerBaseTest.java
upendrakumbham: Sure

build.gradle
```diff
@@ -105,10 +111,10 @@ dependencies {
     testImplementation 'org.hamcrest:java-hamcrest:2.0.0.0'
     testImplementation 'commons-beanutils:commons-beanutils:1.9.3'
 
-    testCompile 'org.junit.jupiter:junit-jupiter-api:5.4.1'
-    testRuntime 'org.junit.jupiter:junit-jupiter-engine:5.4.1'
-    testCompile 'junit:junit:4.12'
-    testRuntime 'org.junit.vintage:junit-vintage-engine:5.4.1'
+    testImplementation "org.junit.jupiter:junit-jupiter-api:$junitJupiterVersion"
+    testImplementation "org.junit.jupiter:junit-jupiter-params:$junitJupiterVersion"
```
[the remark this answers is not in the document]
upendrakumbham: I'm not really sure. Need to double-check.

build.gradle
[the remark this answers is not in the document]
upendrakumbham: Apologies, I take care!

alfonsomunozpomer:
Superseded by https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/166.

PR #169, opened 2021-06-17
Feature/174839971 add new request parameter for umap basing develop

lingyun1010:


upendrakumbham (approved):
All looks good to me. Clarifications and tiny changes if you can.

upendrakumbham (changes requested):
Please commit test cases as well with the @Ignore option. So that tests will be ignored for time being and we can remove the @Ignore option and test after fixing our issue.

lingyun1010:
This one was merged to branch `umap-integration` before.

PR #178, opened 2021-07-23
Chore #178989566 – Remove bottlenecks and improve load time of experiment page

alfonsomunozpomer:
In summary I removed the cell count trip to Postgres and went with `analysedAssays` which is cached with each experiment object. It’s worth mentioning that [there’s a difference of 1 there](https://www.pivotaltracker.com/story/show/178988294), so somewhere we’re forgetting that some objects start at 0... who knows?

The other, more important bit, is removing the retrieval of perplexities because it’s redundant with plot types and options, which is cached now.


lingyun1010 (approved):
LGTM! Nice improvements.

app/src/main/java/uk/ac/ebi/atlas/experimentpage/tabs/ExperimentPageContentService.java
```diff
@@ -67,8 +67,6 @@ public JsonObject getTsnePlotData(String experimentAccession) {
         tsnePlotSettingsService.getExpectedClusters(experimentAccession)
                 .ifPresent(value -> result.addProperty("selectedK", value));
 
-        result.add("perplexities", getPerplexities(experimentAccession));
```
lingyun1010, line 70: Please modify the test for this.

alfonsomunozpomer: I missed it! Will do, many thanks.

alfonsomunozpomer: I leave this as it is, since this is fixed in the upcoming PR with the new cell plots endpoint. Well spotted!

PR #179, opened 2021-07-23
Feature #178989736 – Change “t-SNE plots” to “Cell plots”

alfonsomunozpomer:
We forgot to creat a ticket for this; Irene agreed to this rephrasing, so here it is.


pcm32 (approved):
LGTM, wording is the agreed, thanks.

Issue #162, opened 2021-04-15
Single cell experiment page link broken for the experiment - E-MTAB-7316


upendrakumbham:
Issue: Experiments page has been broken for the experiment:  E-MTAB-7316

[E-MTAB-7316](http://ves-hx-77:8080/gxa/sc/experiments/E-MTAB-7316) - 400 error.

User ticket on RT:  https://helpdesk.ebi.ac.uk/Ticket/Display.html?id=489128

Pivotal reference: https://www.pivotaltracker.com/story/show/177533187

upendrakumbham:
I have started my debugging to find a root cause:

I'm able to reproduce the bug. I could see '400' error code when I try to fetch the experiment: E-MTAB-7316 on `ves-hx-77`(test)

upendrakumbham:
During code debugging, I have seen Solr Exceptions while fetching experiment:

`Class - ExperimentController.java -> endpoint-> "/experiments/{experimentAccession}" method-> showExperimentPage(..)`

In this method, while adding content to the model code is failing.

```
...model.addAttribute(
                "content",
                experimentPageContentSerializer.experimentPageContentForExperiment(experimentAccession, accessKey));
```

upendrakumbham:
I have decided to switch test Solr to point to analytics v3 collection instead of v5 just to prove that collection issue or not.
I have created a new alias on [Test Solr](http://wp-p3s-67:8983/solr/#/~collections) and pointed to the `scxa-analytics-v3` collection from the` scxa-analytics-v5`.

upendrakumbham:
I have tested after pointing test Solr to v3 and found that the experiment page working fine for the experiment: E-MTAB-7316. I have got the `200` status code.

upendrakumbham:
Again I pointed [Test Solr](http://wp-p3s-67:8983/solr/#/~collections) to analytics v5 collection to find out java problematic a piece of code. Here are the details and exceptions traces...

Problematic java code -  
```

Class- CellMetaDataDao.java -> 

method->

// Given a type of metadata and an experiment accession, this method retrieves the value of that metadata for the
    // cells in an experiment
    public ImmutableMap<String, String> getMetadataValues(String experimentAccession,
                                                          String metadataType) {
        // Find matching characteristic name and factor name docs...
 var solrQueryBuilder =
                new SolrQueryBuilder<SingleCellAnalyticsCollectionProxy>()
[...]

PR #185, opened 2021-07-26
Bugfix #177533187 – Log an error if a metadata value is missing

alfonsomunozpomer:
Rather than showing a cryptic 400 error we use the placeholder “Not available” for missing metadata values. I haven’t included a test because:
- We need to use an incorrect condensed SDRF file
- Verifying that the logger adds an error message would mean a mocked logger, a spy, or a somewhat convoluted constructor just for the sake of testing

In summary: testing a side effect as a consequence of an edge case doesn’t seem like the most solid business logic and should happen rarely.

### DIFF
--- a/app/src/main/java/uk/ac/ebi/atlas/experimentpage/metadata/CellMetadataDao.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/experimentpage/metadata/CellMetadataDao.java
@@ -178,7 +178,7 @@ public class CellMetadataDao {
                             // https://www.pivotaltracker.com/story/show/177533187
                             tuple -> {
                                 if (tuple.getStrings(METADATA_VALUE_TARGET_FIELD_NAME) == null ||
-                                    !tuple.getStrings(METADATA_VALUE_TARGET_FIELD_NAME).isEmpty()) {
+                                    tuple.getStrings(METADATA_VALUE_TARGET_FIELD_NAME).isEmpty()) {
                                     LOGGER.error(
                                             "Missing metadata value for {} – {}",
                                             tuple.getString(CELL_ID.name()),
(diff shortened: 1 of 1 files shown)


alfonsomunozpomer:
Fixes https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/162.

app/src/main/java/uk/ac/ebi/atlas/experimentpage/metadata/CellMetadataDao.java
```diff
@@ -166,7 +168,26 @@ public CellMetadataDao(SolrCloudCollectionProxyFactory solrCloudCollectionProxyF
...
+                            // but it’s a situation that we’ve encountered in the past and we want the web app to be
+                            // somewhat lenient about it and show the experiment, rather than show a cryptic 400 error;
+                            // instead we log an error and fill those fields with “Not available”.
+                            // Please see:
+                            // https://ebi-fg.slack.com/archives/C9P2QETJ9/p1624354966014500
+                            // https://ebi-fg.slack.com/archives/C800ZEPPS/p1624362599003900
+                            // https://www.pivotaltracker.com/story/show/177533187
+                            tuple -> {
+                                if (tuple.getStrings(METADATA_VALUE_TARGET_FIELD_NAME) == null ||
+                                    !tuple.getStrings(METADATA_VALUE_TARGET_FIELD_NAME).isEmpty()) {
```
haideriqbal, line 181: Wouldn't this condition result in the correct Cell Ids being logged as errors? Say for example, `cell_id` is `not null` and its name is `xyz` then this `if` condition would evaluate to true. As it has a `!` before `isEmpty()`.

Or am I missing something here?

haideriqbal (changes requested):
Not much, just one comment! Otherwise LGTM!

haideriqbal (approved):
All Good!

Issue #181, opened 2021-07-23
Experiment page is not landing by default on inferred cell type ontology labels as it should


pcm32:
This is a feature that was introduced in the last release, on landing to an experiment the colouring should be based on the Inferred cell type ontology labels if available. Currently is preferring a clustering value. Clustering colouring should always be at the end of the priority list for coloring. This applies to both tSNE and UMAP.

See correct behaviour in www, for instance for https://www.ebi.ac.uk/gxa/sc/experiments/E-MTAB-9067/results/tsne

haideriqbal:
Closing this issue as it has been fixed here https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/186

PR #186, opened 2021-07-29
Select inferred cell types by default

haideriqbal:
This PR is related to another PR here: https://github.com/ebi-gene-expression-group/atlas-components/pull/84

The idea is to select inferred cell types by default if they exist in color by dropdown 

This PR solves the issue mentioned here: https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/181


app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
```diff
@@ -308,7 +317,12 @@ TSnePlotViewRoute.propTypes = {
     value: PropTypes.string.isRequired
   }).isRequired).isRequired,
   selectedK: PropTypes.number,
-  anatomogram: PropTypes.object.isRequired
+  anatomogram: PropTypes.object.isRequired,
+  initialCellTypeValues: PropTypes.array
+}
+
+TSnePlotViewRoute.defaultProps = {
```
pcm32, line 324: Do we need an equivalent here for UMAP or this is just an unfortunate name being kept for this object and this applies as well to the UMAP dimred with nothing in addition needed?

pcm32:
Please @haideriqbal quote here which issue this PR is fixing, thanks.

app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
haideriqbal: Its just the component name we have which we didn't change. We do not need anything additional for UMAP.

haideriqbal:
Once this PR is accepted then the other PR https://github.com/ebi-gene-expression-group/atlas-components/pull/84 will become redundant and we can close that.

app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
```diff
@@ -308,7 +310,12 @@ TSnePlotViewRoute.propTypes = {
     value: PropTypes.string.isRequired
   }).isRequired).isRequired,
   selectedK: PropTypes.number,
-  anatomogram: PropTypes.object.isRequired
+  anatomogram: PropTypes.object.isRequired,
+  initialCellTypeValues: PropTypes.array
```
alfonsomunozpomer, line 314: Array of what? I think you can be a bit more specific.

app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
```diff
@@ -42,22 +42,24 @@ class TSnePlotViewRoute extends React.Component {
         }
      ]
 
+    let cellType = _first(_intersection(_map(this.props.metadata,`label`), this.props.initialCellTypeValues))
```
alfonsomunozpomer, line 45: Why `let` and not `const`? Also, is the value of this variable a cell type? I think this name is misleading.

alfonsomunozpomer (commented):
I can see that this solution would work, but I’m on the fence of approving it because I don’t know if this supersedes https://github.com/ebi-gene-expression-group/atlas-components/pull/84 or the changes in the component is required. Be aware that the latter PR introduces a bug in the component behaviour.

app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
haideriqbal: Makes sense will update!

app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
haideriqbal: Makes sense we can use `const` as we do not want to change the variable value later so a better safer option.

Regarding, name I assumed that it's returning different types of cell type values so that's why. Let me think of another word.

haideriqbal:
> I can see that this solution would work, but I’m on the fence of approving it because I don’t know if this supersedes [ebi-gene-expression-group/atlas-components#84](https://github.com/ebi-gene-expression-group/atlas-components/pull/84) or the changes in the component is required. Be aware that the latter PR introduces a bug in the component behaviour.

By approving this PR we will move the functionality to the backend from the frontend. At the moment frontend component do not have any such functionality and we don't have to change anything in the frontend component to make it work. So in my opinion, it is safe to say that this PR supersedes the one mentioned because the changes introduced in PR-84 caused the bug otherwise the component is working fine on its own.

alfonsomunozpomer (approved):
All good!

PR #188, opened 2021-08-04
Bugfix #179005671 – All cells are bucketed in a not defined bucket when plotting by metadata

alfonsomunozpomer:
It all started with an innocent comment: let’s go back to one endpoint for *k* values and another for metadata categories, because the two endpoints had been merged into one.

I decided then to create a new package with a new endpoint that represented the new reality: we no longer have *t-SNE plots* but *Cell plots*. I moved functionality, bit by bit, from the old JSON endpoints to the new ones. There’s a new way to query Postgres JSON fields using the operator `@>` which allows us to use a natural JSON representation of the parameters from a map. Check the class `CellPlotType` and the method in the controller which parses the plot options from the request parameters. Maybe having a class which represents a selected plot method plus the options might make things a bit better, but I’m not sure. I decided to stop modelling problem domain classes with `CellPlotType`. Just notice that adding any new plot type plust its options there will suffice to make the back end work, nothing else would be needed.

Then I thought that we had this bug because we had no way to test the application because our fixtures were broken. That’s when I decided to revisit the fixture scripts and make them better. And I added a set of new fixtures. Currently both old and new are used, but as we move forward and we test and review code we should change all tests to use the new ones. As a consequence I changed some tests to make them more robust (or correct).

I changed stuff in `buildSrc` and Docker to run tests like civilised people (we’re not animals!). Check the `README.md` which contains documentation on this. 

Lastly I rebuilt the dev bundles and used the latest version of `scxa-tsne-plot` in the `experiment-page` bundle.

This PR is accompanied by https://github.com/ebi-gene-expression-group/atlas-web-core/pull/65.


app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotType.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.Optional;
+
+public class CellPlotType {
+    private final static ImmutableMap<String, ImmutableSet<String>> PLOT_OPTIONS =
+            ImmutableMap.of(
+                    "tsne", ImmutableSet.of("perplexity"),
+                    "umap", ImmutableSet.of("n_neighbors"));
+
+    public final static String DEFAULT_PLOT_METHOD = "umap";
```
haideriqbal, line 14: A minor comment, but wouldn't it be better to name it as `DEFAULT_CELL_PLOT_TYPE` just to make it consistent with the class name.

app/src/test/java/uk/ac/ebi/atlas/download/ExperimentFileLocationServiceIT.java
```diff
@@ -75,27 +75,25 @@
...
+				new ClassPathResource("fixtures/202108/scxa_cell_group_membership.sql"),
+				new ClassPathResource("fixtures/202108/scxa_cell_group_marker_genes.sql"),
+				new ClassPathResource("fixtures/202108/scxa_cell_group_marker_gene_stats.sql"));
 		populator.execute(dataSource);
 	}
 
 	@AfterAll
 	void cleanDatabaseTables() {
-		ResourceDatabasePopulator populator = new ResourceDatabasePopulator();
+		var populator = new ResourceDatabasePopulator();
```
haideriqbal, line 90: 💯

haideriqbal (approved):
Just one minor comment. Otherwise, all good! Ran all the test cases and they passed as well! 🎖️

app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.Map;
+import java.util.LinkedHashMap;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static java.util.function.Function.identity;
+import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.mapping;
+import static java.util.stream.Collectors.toList;
+
+@Component
```
upendrakumbham, line 20: I know it's minor. But it's a good time to rename as we changed the service name.

app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/JsonCellPlotController.java
```diff
@@ -0,0 +1,110 @@
...
+
+        // Check that no param is missing
+        var requiredParametersBuilder = ImmutableMap.<String, Integer>builder();
+        for (var requiredParameter : requiredParameters) {
+            if (!requestParams.containsKey(requiredParameter)) {
+                throw new IllegalArgumentException("Missing parameter " + requiredParameter);
+            } else {
+                try {
+                    requiredParametersBuilder.put(
+                            requiredParameter, Integer.parseInt(requestParams.get(requiredParameter)));
```
upendrakumbham, line 43: I had a feeling by looking at these validations that would it be possible to move these validations to any helper class of that package. We can ignore this for the time being and pick this one in the future clean code improvements :)

lingyun1010 (approved):
Looks good for me!

app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotType.java
alfonsomunozpomer: I chose that name to align it with the request parameter `plotMethod` in the class `JsonCellPlotController`. I modelled a plot type as two values: a plot mehotd and a set of parameters. E.g. a UMAP plot type has a plot method `umap` and a set of parameters with a single element, `n_neighbors`. Put it another way, a plot method is the identifier of a plot type.

app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/JsonCellPlotController.java
alfonsomunozpomer: It makes sense. I restrained from disgreagating this into another class because the controller is the only place where this logic is used. If at some point other classes need to use this piece of code we might as well put it in a separate class. This also helps to test it since the logic is non-trivial.

PR #191, opened 2021-09-01
Chore #179298051 – Test and build in Jenkins

alfonsomunozpomer:
Quite a few things to unpack here.

The first thing: *DO NOT*, I repeat and emphasise, *DO NOT* try to follow the commit history here. There are many revertions and a lot of trial and error. Just focus on the final changes to keep your sanity.

The meaty parts are `Jenkinsfile` and `jenkins-k8s-pod.yaml`. The latter is a pod with an OpenJDK container to run Gradle, a Postgres container that the integration tests use as DB, and a few persistent volumes to make everything work. The volumes contain the [test experiment datasets](http://ftp.ebi.ac.uk/pub/databases/microarray/data/atlas/test/scxa/filesystem/) and the [`bioentity_properties` directory](http://ftp.ebi.ac.uk/pub/databases/microarray/data/atlas/bioentity_properties/). There are two other volumes to speed up build times: one that contains all the custom packages, intended as a fallback if the Artifactory repo is offline, and another one that keeps the Gradle cached objects (i.e. `/root/.gradle`) to avoid re-downloading dependencies on every build. In `Jenkinsfile` you can see all the testing steps except the IT phase of `atlas-web-core`, which needs a volume with the bulk testing experiments and bulk collections. Unfortunately, we developed `atlas-web-core` before Single Cell Expression Atlas, and because of this the library is tested against bulk. Once we have bulk testing in Jenkins we can enable ITs there.

Other interesting changes:
- If we run Flyway as a container, it will run the DB migrations and then terminate the container. Jenkins build pods terminate whenever *any* container does, so rather than try to run some arbitrary `sleep` in combination with the default entry point I moved it to Gradle.
- The Postgres container only allows to define a single DB in the environment variable `POSTGRES_DB`, but we need two: `gxa` for `atlas-web-core` and `scxa` for `app`. To solve it I configured the container to use a single, generic, `postgres` database and use two schemas. It’s really good that Flyway initialised the schemas if they don’t exist and you can add the `currentSchema` as a request parameter to `jdbcUrl`. In the end, it works as a different scope.
- Some tests have been changed to pass with [the latest iteration of our Solr collection, `scxa-analytics-v6`](https://github.com/ebi-gene-expression-group/index-scxa/pull/29).
- If the build is run on the `develop` or `main` branches a WAR file is generated with the bundles, which are now Git-ignored. However, and this is important, you need to update the bundles locally to verify that they work. Jenkins will build them according to your `package.json` and `package-lock.json` files. There is a new script, `update-compile-front-end-packages.sh`, that automates all this. The script updates [`@ebi-gene-expression-group` packages](https://www.npmjs.com/org/ebi-gene-expression-group) to the latest published versions, so be careful: if a package isn’t published it won’t be included in the build.

In the end, this is the result:
image


haideriqbal (approved):
Code wise, it looks good to me! How are we going to test if Jenkin is working fine or not? Do we need to create a local k8s cluster?

upendrakumbham (approved):
I have gone through & understood most of the pieces. Precise context description to make it easy for the reviewers to review the PR.

lingyun1010 (approved):
Looks good for me.

alfonsomunozpomer:
Thanks everybody for taking the time to look at this. It’s going to make our testing easier.

PR #196, opened 2021-09-24
Chore #179298051-#177741644 – Test and build web app on Jenkins

alfonsomunozpomer:
After many trials and errors we got it. See the results: http://45.88.81.188:30004/jenkins/view/Expression%20Atlas%20web%20apps/job/Single%20Cell%20Expression%20Atlas/job/k8s-jenkins/

At the moment you’ll see that `FileDownloadControllerWIT` is explicitly disabled. We’ll eventually need to review this test.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,6 +1,7 @@
 pipeline {
   agent {
     kubernetes {
+      cloud 'atlas-analysis'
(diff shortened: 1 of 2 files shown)


upendrakumbham (approved):
Pure Configuration :). LGTM!

PR #198, opened 2021-09-27
Feature/178928604 enable access for private experiments rebase

upendrakumbham:
This feature allows users to see private experiments if they know the access key in the single cell.
Companion PR: https://github.com/ebi-gene-expression-group/atlas-web-core/pull/70


alfonsomunozpomer (approved):
All looks good and the tests pass. I think this is good to merge.

PR #212, opened 2021-10-14
Feature #179903156 – Merge same marker gene to one row

alfonsomunozpomer:
We suffered a regression after adding the cell group tables to our DB. Before we only picked the marker gene in the group/cluster in which it had the lowest p-value. The merge now is done on the back end code, at serialisation time, since it’s a visualisation decision.

I also added some volumes to cache Gradle artifacts to speed up builds.


upendrakumbham (approved):
LGTM!

PR #217, opened 2022-01-17
Refactor to correctly select colourBy category

haideriqbal:
TSne plot was not being correctly coloured when going through a URL. Details are in this [story](https://www.pivotaltracker.com/story/show/180331840).

The reason was that by default if there are cell types in the metadata prop then `metadata` was chosen as a default category to colour plots without checking the params passed in the URL. The issue has been resolved by checking URL params first to see if colourBy param is an integer or not and then selecting proper colourBy category.

### DIFF
--- a/app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
+++ b/app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
@@ -43,6 +43,7 @@ class TSnePlotViewRoute extends React.Component {
      ]
 
     const cellTypeValue = _first(_intersection(_map(this.props.metadata,`label`), this.props.initialCellTypeValues))
+    const search = URI(this.props.location.search).search(true)
 
     this.state = {
       selectedPlotType: plotTypeDropdown[0].plotType.toLowerCase(),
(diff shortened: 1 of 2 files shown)


app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
```diff
@@ -53,7 +54,7 @@ class TSnePlotViewRoute extends React.Component {
       selectedColourBy: cellTypeValue ? cellTypeValue.toLowerCase() : this.props.ks[Math.round((this.props.ks.length -1) / 2)].toString(),
       highlightClusters: [],
       experimentAccession: this.props.experimentAccession,
-      selectedColourByCategory: cellTypeValue ? `metadata` : `clusters`
+      selectedColourByCategory: !isNaN(search.colourBy) ? `clusters` : cellTypeValue ? `metadata` : `clusters`
```
alfonsomunozpomer, line 57: Maybe `Number.isSafeInteger(search.colourBy)` is more readable? As usual there’s more than one way to do this and you decide, it’s a matter of taste really.

alfonsomunozpomer (changes requested):
Could you please add a test?

app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
haideriqbal: I believe we will need to add an extra function to the suggested solution as `search.colourBy` is a string and for `Number.isSafeInteger` it returns `false` for anything which is not a number value. So we have two options here either we do something like this `Number.isSafeInteger(Number(search.colourBy))` to convert `search.colourBy` to a number type, and then perform `isSafeInteger` check or keep the solution as it is. I personally like the latter. Let me know what you think?

haideriqbal: https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_issafeinteger2

alfonsomunozpomer: Yes, you’re totally right. I forgot about the type coertion stuff. Leave as it is then.

PR #220, opened 2022-02-04
Fix cell count discrepancy 

haideriqbal:
Created a new Dao class to retreive cell counts for the database. Relevant web-core [PR](https://github.com/ebi-gene-expression-group/atlas-web-core/pull/94).


app/src/test/java/uk/ac/ebi/atlas/experimentpage/TSnePlotSettingsServiceIT.java
```diff
@@ -115,6 +119,16 @@ void getPerplexitiesForValidAccession() {
                 .doesNotHaveDuplicates();
     }
 
+    @Test
+    void getCellCountForValidAccession() {
+        Integer result =
```
upendrakumbham, line 124: Use 'var'

app/src/test/java/uk/ac/ebi/atlas/experimentpage/TSnePlotSettingsServiceTest.java
```diff
@@ -214,6 +218,17 @@ void noPerplexities() {
         assertThat(subject.getAvailablePerplexities(EXPERIMENT_ACCESSION)).isEmpty();
     }
 
+    @Test
+    @DisplayName("Valid cell count from database")
+    void validCellCount() {
+        Integer expectedCellCount = 100;
+        when(experimentCellCountDaoMock.fetchNumberOfCellsByExperimentAccession(EXPERIMENT_ACCESSION))
```
upendrakumbham, line 225: You can use hardcoded `100` instead of an extra local variable as we are using it in one place.

haideriqbal: We are using `expectedCellCount` at two places in this test case, and if we hard code `100` then it will need to be done at two places. So imo that would affect the maintainability of the code in the future if we need to change this value then it would need to be changed in two places rather than in just one place.

upendrakumbham: yeah, agree with you :).

upendrakumbham (dismissed):
All Good. Please merge. Verified both places, cell counts are in sync as expected.

upendrakumbham (dismissed):
All looks good to me!

PR #229, opened 2022-03-08
feature/181415740-Aggregate metadata and gene id suggestions

upendrakumbham:
- Aggregate metadata and gene id suggestions
- Generic endpoint implementation for both bioentity and analytic suggesters and concat both streams as a single response.


app/src/main/java/uk/ac/ebi/atlas/search/AutocompleteController.java
```diff
@@ -85,6 +86,15 @@ public String fetchMetaDataSuggestions( @RequestParam(value = "query") String qu
                 query,species.split(","))));
     }
 
+    @GetMapping(value = "/json/suggestions/gene_search",
```
haideriqbal, line 89: Can you please explain why we need this additional endpoint? When looking at the code, the logic and params of `fetchGeneIdSuggestions` and `fetchGeneIdAndMetadataSuggestions` is exactly the same. We can use the same endpoint to achieve the same result. Unless I'm missing something here.

app/src/test/java/uk/ac/ebi/atlas/search/AutocompleteControllerWIT.java
```diff
@@ -73,57 +73,76 @@ void setUp() {
...
     void geneIdSuggestions() throws Exception {
-        this.mockMvc
-                .perform(get("/json/suggestions/gene_ids").param("species", "Homo sapiens").param("query", "asp"))
-                .andExpect(status().isOk())
-                .andExpect(content().contentType(MediaType.APPLICATION_JSON_UTF8))
-                .andExpect(jsonPath("$[*].label", hasSize(greaterThanOrEqualTo(1))))
-                .andExpect(
-                        jsonPath(
-                                "$[*].label",
-                                everyItem(in(BIOENTITY_PROPERTY_NAMES.stream().map(p -> p.label).collect(toList())))))
```
haideriqbal, line 97: Why were these statements removed from the test?

app/src/test/java/uk/ac/ebi/atlas/search/AutocompleteControllerWIT.java
```diff
@@ -73,57 +73,76 @@ void setUp() {
...
+          .andExpect(status().isOk())
+          .andExpect(content().contentType(MediaType.APPLICATION_JSON_UTF8));
+    }
+
+    @Test
+    void canSeeSuggestionsForTheGeneId() throws Exception {
+        this.mockMvc.perform(get("/json/suggestions/gene_search")
+          .param("species", "")
+          .param("query", "asp"))
+          .andExpect(status().isOk()).andExpect(content()
```
haideriqbal: Please check the formatting settings of your IDE. Formatting correctly enhances readability of code.

Screenshot 2022-04-08 at 16 32 53

haideriqbal (dismissed):
Overall it looks good. Please see the comments and check the formatting settings of your IDE as it has changed the formatting of the older code as well.

For future reference, please add a bit of context to the PR description about what is the issue/feature is and how we are implementing it. This gives the reviewer a bit of background when reviewing the code and why certain things have changed.

app/src/main/java/uk/ac/ebi/atlas/search/AutocompleteController.java
upendrakumbham: True, this additional new endpoint replica of the `gene_id` endpoint. The future plan is to eliminate this `gene_id `endpoint and use only the `gene_search` endpoint as `gene_id` endpoint is too specific. As you knew already we introduced metadata search as well. That's why @alfonsomunozpomer suggested to implement a generic endpoint. We can replace the `gen_id` endpoint with `gene_search` on the front end.

app/src/test/java/uk/ac/ebi/atlas/search/AutocompleteControllerWIT.java
upendrakumbham: Now we are getting aggregated JSON response - `metadata+geneid `suggestions. That matcher condition is not meeting.

app/src/test/java/uk/ac/ebi/atlas/search/AutocompleteControllerWIT.java
upendrakumbham: Agree. It's a big puzzle for me, frequently I'm facing this issue. Let me look at IDE settings.

app/src/main/java/uk/ac/ebi/atlas/search/AutocompleteController.java
haideriqbal: If I understand correctly, the functionality of both endpoints is the same at the moment and it's just the name of the endpoint which is different. So why can't we change the endpoint to the `gene_search` now in the front-end? and have only one endpoint in the backend?

upendrakumbham: I'm not sure, @alfonsomunozpomer suggested this approach so that we can update the front end in a feasible time. I think if we replace it suddenly with the new one,  it might impact other components which are dependent only on the backend.

haideriqbal: ahh okay now it makes sense. Yeah I guess then its fine we can update the front-end endpoints in meanwhile.

app/src/test/java/uk/ac/ebi/atlas/search/AutocompleteControllerWIT.java
upendrakumbham: Done - code reformat.

alfonsomunozpomer (approved):
LGTM. Merge away!

PR #231, opened 2022-03-10
Feature #181187761 – Render information banner from external file

alfonsomunozpomer:
@upendrakumbham please have a look at this way of rendering the banner. I know this is right after you fixed it for Single Cell but I had it almost finished last week before I went on sick leave... sorry for not pushing the WIP!

I think this way of doing things give us our envisioned scenario in which curators can just edit https://github.com/ebi-gene-expression-group/ebi-gene-expression-group.github.io/blob/main/scxa-motd.md and display a relevant message, or leave the file blank if we don’t want to show anything at all.

### DIFF
--- a/app/src/main/javascript/bundles/atlas-information-banner/index.js
+++ b/app/src/main/javascript/bundles/atlas-information-banner/index.js
@@ -1,10 +1,22 @@
 import React from 'react'
 import ReactDOM from 'react-dom'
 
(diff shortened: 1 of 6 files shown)


upendrakumbham (approved):
Nice feature, save us from rework. All looks good to me. I tried on the local machine, working fine. Gone through the atlas reach fetch loader component usage.

PR #241, opened 2022-05-17
Add a URL to enable import new experiment

lingyun1010:
Add a URL that we can hit to force-evict public experiments-related caches and cache newly-added public experiments

Through the use of the atlas-web-sc-cli we are now able to load experiments completely independent of the tomcat process running SC Expression Atlas. However, once loaded into the database, expDesign, solr, magetab, etc the web application will only see the new experiment on wwwdev on a restart. It would be useful to have an endopoint to poke with an accession, so that the web application becomes aware of that experiment and loads it into its memory structures.

To clarify, this doesn't need to trigger a load to the database or do the expDesign file, but simply do the same with a specific experiment as it is done with all available experiments on restart.


app/src/main/java/uk/ac/ebi/atlas/experimentimport/admin/SingleCellExperimentAdminController.java
```diff
@@ -27,4 +35,13 @@ public SingleCellExperimentAdminController(DataFileHub dataFileHub,
     public String fooBar() {
         return "";
     }
+
+    @GetMapping(value = "/{experimentAccession}")
+    public void updateSingleCellExperiment(
+            @PathVariable String experimentAccession,
+            @RequestParam(value = "isPrivate") Boolean isPrivate,
+            ScxaExperimentCrud scxaExperimentCrud) {
+        UUID accessKeyUUID = scxaExperimentCrud.createExperiment(experimentAccession, isPrivate);
```
upendrakumbham, line 44: Hi @lingyun1010 , can you please clarify me? 

`createExperiment ` method,  serve these things:

1. Clear the existing cache and allows the application to cache the latest data sets
2. Reads experiment from the DB by experiment accession
3. Creates a new experiment in the DB if the access key is empty 
4.  Updates existing experiment if the access key is available

what could be the reason to call this method here? only to clear cache? or only update an experiment or both?

app/src/main/java/uk/ac/ebi/atlas/experimentimport/admin/SingleCellExperimentAdminController.java
```diff
@@ -27,4 +35,13 @@ public SingleCellExperimentAdminController(DataFileHub dataFileHub,
...
+        LOGGER.info("Import experiment successfully, accessKeyUUID: {}", accessKeyUUID.toString());
+    }
```
upendrakumbham, line 46: Please update this logger message accordingly, does it create a new experiment or update experiment & clear cache?

app/src/main/java/uk/ac/ebi/atlas/experimentimport/admin/SingleCellExperimentAdminController.java
lingyun1010: We would like to provide this URL to update or add new experiments to the cache.

app/src/main/java/uk/ac/ebi/atlas/experimentimport/admin/SingleCellExperimentAdminController.java
lingyun1010: We import an experiment. If this experiment is existed in the cache, we'll update it, if not, we'll add it. So `import` will be a proper logger message and I take the reference from where we call `createExperiment` in the existing code.

upendrakumbham: After our chat, here is a minor comment to update the codebase:

Please rename controller method name as a `createOrUpdateExperiment()` because `createExperiment()` from crud is doing create/update.

We can remove the logger in the controller if it passes the same info from the DAO class.

If you want to keep it -  update it like below 

```
LOGGER.info("Imported/updated experiment successfully, accessKeyUUID: {}", accessKeyUUID.toString());

```
All good.

upendrakumbham (dismissed):
LGTM! Remove logger before the merge as it gives the same info as DAO.

upendrakumbham (approved):
Added a minor comment, please have a look and merge away.

PR #243, opened 2022-05-30
Bugfix #182315539 - Add organism in payload field

alfonsomunozpomer:
To filter suggestions by species rather than using the `suggest.cfq` field we want to do it in the back end (see https://github.com/ebi-gene-expression-group/index-scxa/pull/37 for a more in-depth explanation).

The species argument has been removed from the DAO and a few improvements have been made here and there.

Accompanied by https://github.com/ebi-gene-expression-group/atlas-web-core/pull/108.


upendrakumbham (approved):
All looks good to me.

PR #244, opened 2022-05-31
Feature #181415976 – Redirect metadata terms to cell type wheel page

alfonsomunozpomer:
Not much to say, we parse the `queryObject` (which should come from the variable `query` rather than `geneQuery` to make it consistent, but since needs changes from the front end I decided to leave it be), and redirect to the cell type wheel page if the term is of category `metadata`.


lingyun1010 (approved):
Looks all good for me! Please merge!

PR #253, opened 2022-06-09
Bugfix #182329349 – Serialize bioentity and metadata suggestions in a single list

alfonsomunozpomer:
So this is the way to do it: just a PR with the updated module.


upendrakumbham (approved):
LGTM!

PR #262, opened 2022-06-30
Feature #182401190 – Reword main search form

### LINKED ISSUE #246: Metadata search: Hint to the users that conditions can be searched now
In the main search we currently see:

image

I would expect this to hint currently that not only Gene IDs or symbols can be searched for, but also conditions and cell types. So maybe "Gene ID, gene symbol, cell type or condition/factor" and possibly expand examples below (I would love to see for instance as example "epithelial cell of lung") which should take us to datasets that also have anatomograms. Maybe the curators can suggest more example to put below.

alfonsomunozpomer:
Remove label and add metadata search examples.

Solves #246 and https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/262.

### DIFF
--- a/app/src/main/webapp/WEB-INF/jsp/tiles/views/home/search.jsp
+++ b/app/src/main/webapp/WEB-INF/jsp/tiles/views/home/search.jsp
@@ -1,5 +1,7 @@
 <%--@elvariable id="resourcesVersion" type="String"--%>
 <%--@elvariable id="geneId" type="String"--%>
+<%--@elvariable id="resourcesVersion" type="String"--%>
+<%--@elvariable id="geneId" type="String"--%>
 
 <%@ page contentType="text/html;charset=UTF-8" %>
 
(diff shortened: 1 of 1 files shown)


app/src/main/webapp/WEB-INF/jsp/tiles/views/home/search.jsp
```diff
@@ -1,5 +1,7 @@
 <%--@elvariable id="resourcesVersion" type="String"--%>
 <%--@elvariable id="geneId" type="String"--%>
+<%--@elvariable id="resourcesVersion" type="String"--%>
```
ke4: Are lines 3 and 4 the duplicate of lines 1 and 2?
If yes, then could we remove them, please?

app/src/main/webapp/WEB-INF/jsp/tiles/views/home/search.jsp
```diff
@@ -15,41 +17,54 @@
...
-            text: 'MGI:98354 (MGI ID)',
-            url: '${pageContext.request.contextPath}/search?mgi_id=MGI:98354'
-          },
-          {
-            text: 'FBgn0004647 (FlyBase ID)',
-            url: '${pageContext.request.contextPath}/search?flybase_gene_id=FBgn0004647'
-          }
-        ]
-      }, 'search-form')
+            autocompleteLabel: ``,
```
ke4: Should these be single quotes? I am not sure.

```suggestion
            autocompleteLabel: '',
```

ke4 (changes requested):
Hey, I added 2 tiny suggestions to consider.
Could you have a look at them, please?
Otherwise, it looks fine.
Thanks

app/src/main/webapp/WEB-INF/jsp/tiles/views/home/search.jsp
alfonsomunozpomer: In our React packages we use [backtick strings](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) but in our JSPs we should use old-style JavaScript, with either double or single quotes. We chose the latter. So yes!

app/src/main/webapp/WEB-INF/jsp/tiles/views/home/search.jsp
alfonsomunozpomer: Oh, I missed those. Thanks!

ke4 (approved):
LGTM 👍

PR #264, opened 2022-06-30
Bugfix #261 – Gene search with any species sets request param `species=undefined`

### LINKED ISSUE #261: Gene search with “Any” species sets request parameter `species=undefined` and returns no results

alfonsomunozpomer:
Create links in the browser History using URI rather than by manually wrangling strings.


lingyun1010 (approved):
Smart way to fix this URL issue.

PR #270, opened 2022-07-07
Feature #182401363 – Add a request parameter to the suggestions endpoint to omit metadata terms

alfonsomunozpomer:
I created a new `SuggestionsController` with the following features:
- Go with the default Spring annotations, which are sensible, so we can remove a noticeable amount of boilerplate code
- Reworked the paths because it was all getting very confusing!
- I noticed that `json/suggestions/meta_data` and `json/suggestions/gene_search` are redundant
- Renamed the paths to be consistent with the convention of splitting words with hyphens rather than underscores
- Removed the species from the new class because it’s not suggesting anything, in time we can move this to a separate endpoint

This means that we’ll need to update our front end components to use this new URL, but we should keep the old one until we’re sure we’re not using it at all. The suggestions endpoint is being called from several places (main tab, gene search tab and experiment page, mayber other places I can’t remember now). Search for `json/suggestions` in `atlas-components` and the `javascript` dir to be sure.

### DIFF
--- a/app/src/main/java/uk/ac/ebi/atlas/search/SuggestionsController.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/search/SuggestionsController.java
@@ -1,0 +1,89 @@
+package uk.ac.ebi.atlas.search;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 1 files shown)


ke4:
We don't need specific suggesters for different views. We are using the same suggesters everywhere.

PR #273, opened 2022-07-08
Feature ##182139622 (Issues#271) Add marker gene endpoint

### LINKED ISSUE #271: Add an endpoint to check if a gene search is a marker gene in any experiments
The parameters in the request should be the same as the gene search endpoint and return a boolean value.

/json/search/marker-genes?

ke4:
Changes:
- Add a new endpoint to check if a gene is a marker gene
- Move the gene query building code from the controller class to the GeneIdSearchService.
- Move other common logic with gene search and marker gene search endpoints to private methods
- Also refactored the gene search endpoint to use the refactored logic
- Add a couple of helper scripts:
  - executing a simple unit test
  - executing all unit tests
  - debugging a specific unit test
  - stop and remove all containers regarding to testing


app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -16,15 +16,12 @@
 import uk.ac.ebi.atlas.model.experiment.singlecell.SingleCellBaselineExperiment;
 import uk.ac.ebi.atlas.search.geneids.GeneIdSearchService;
 import uk.ac.ebi.atlas.search.geneids.GeneQuery;
-import uk.ac.ebi.atlas.solr.bioentities.BioentityPropertyName;
 import uk.ac.ebi.atlas.species.SpeciesFactory;
 import uk.ac.ebi.atlas.trader.ExperimentTrader;
 import uk.ac.ebi.atlas.utils.StringUtil;
 
+import java.util.*;
```
alfonsomunozpomer: We discourage the use of wildcard imports. This was probably IDEA’ s doing. Can you change your settings? You can set it to a high enough number so that you’ll never get `*`:
image

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -68,54 +62,18 @@ public JsonGeneSearchController(GeneIdSearchService geneIdSearchService,
...
 
         // We found expressed gene IDs, let’s get to it now...
-        var geneIds2ExperimentAndCellIds =
-                geneSearchService.getCellIdsInExperiments(geneIds.get().toArray(new String[0]));
-
-        var expressedGeneIdEntries =
-                geneIds2ExperimentAndCellIds.entrySet().stream()
-                        .filter(entry -> !entry.getValue().isEmpty())
-                        .collect(toList());
+        List<Map.Entry<String, Map<String, List<String>>>> expressedGeneIdEntries =
```
alfonsomunozpomer: This is the sort of place where `var` shines. :)

debug_single_test.sh
```diff
@@ -0,0 +1,22 @@
...
+
+export TEST_CASE_NAME=$1
+
+echo "Testing ${TEST_CASE_NAME}"
+
+export ATLAS_DATA_PATH=~/dev/gxa/data
+export POSTGRES_HOST=scxa-postgres-test
+export POSTGRES_DB=gxpscxatest
+export POSTGRES_USER=scxa
+export POSTGRES_PASSWORD=scxa
```
alfonsomunozpomer: I don’t hink we need to `export` all this. Setting the variables would be enough, right?

Also, the value of `ATLAS_DATA_PATH` will likely be different across machines. How about the more portable 
`ATLAS_DATA_PATH=${ATLAS_DATA_PATH:-"~/dev/gxa/data"}`

Maybe we could extend that pattern to all the other variables?

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -179,6 +137,68 @@ public String search(@RequestParam MultiValueMap<String, String> requestParams)
...
+        }
+
+        if (geneIds.get().isEmpty()) {
+            return GSON.toJson(
+                    ImmutableMap.of(
+                            "results", ImmutableList.of(),
+                            "reason", "No expression found"));
+        }
+
+        return null;
```
alfonsomunozpomer: This doesn’t look like a good idea. I think a safer option is to return an `Optional` and `return Optional.empty()` rather than `return null`. The return value and how you deal with it should change too. We try to avoid any explict `null`s in our code. In fact you may have seen some `package-info.java` files around with, e.g.:
```java
@NonNullApi
package uk.ac.ebi.atlas.trader;

import org.springframework.lang.NonNullApi;
```

This helps your IDE to identify those `return` statements and possibilities of passing/returning `null` and warn about it.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -179,6 +137,68 @@ public String search(@RequestParam MultiValueMap<String, String> requestParams)
                         "checkboxFacetGroups", ImmutableList.of(MARKER_GENE.getTitle(), ORGANISM.getTitle())));
     }
 
+    @RequestMapping(value = "/json/search/marker-genes",
```
alfonsomunozpomer: Rather than “overloading” the `/json/search` let’s add a new `/json/gene-search` path. Initially we only searched genes, so there was no ambiguity about the endpoint. Now, since we can also search metadata terms, I think it’s better if we split it.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -179,6 +137,68 @@ public String search(@RequestParam MultiValueMap<String, String> requestParams)
...
+
+        return geneIds2ExperimentAndCellIds.entrySet().stream()
+                        .filter(entry -> !entry.getValue().isEmpty())
+                        .collect(toList());
+    }
+
+    private boolean isRequestParamsEmpty(MultiValueMap<String, String> requestParams) {
+        return requestParams == null
+                || requestParams.size() == 0
+                || (requestParams.containsKey("q") && Objects.equals(requestParams.getFirst("q"), ""));
```
alfonsomunozpomer, line 178: Won’t this violate the contract in the comment?
```java
        // We support currently only one query term; in the unlikely case that somebody fabricates a URL with more than
        // one we’ll build the query with the first match.
```

Meaning that something like `?q=&q=some+valid+query` would fail, when it should search for "some valid query". Maybe we could add a test to describe the expected behaviour?

alfonsomunozpomer (changes requested):
A good chunk of work! The scripts are going to be very useful and refactoring the `geneQuery` methods will also dry things out. Besides the observations in the PR, two issues about style:
- `var` is your friend! Let’s use it as much as possible. :)
- Could you please rename the scripts with kebab-case instead of snake_case to be consistent with the other scripts in the repo?

Last (and very minor), how about moving `SearchTestUtil` as a private method in `GeneIdSearchServiceTest`? Since it’s only used once and in a single test it’s less clutter. Unless you see immediate places where we can reuse it (I haven’t personally looked closer, but if that’s the case no problem leaving it as a separate class).

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: Ah, I already changed this setting a couple of weeks ago but forgot to check all the files I edited before. Sorry about that.
Imports have been reset and cleaned.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: Good spot. Thanks.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: Totally agree. Did a bit of refactoring around this code. I hope it is better.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: OK, I changed it, but I am still not convinced that we need the `/json` prefix in the path.
I left it there for now.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: Refactored the code and add a couple of more tests to some of the classes.

debug_single_test.sh
ke4: Done

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -68,54 +64,17 @@ public JsonGeneSearchController(GeneIdSearchService geneIdSearchService,
...
-        var queryTerm = requestParams.getFirst(category);
-
-        var geneQuery = category.equals("q") ?
-                species
-                        .map(_species -> GeneQuery.create(queryTerm, _species))
-                        .orElseGet(() -> GeneQuery.create(queryTerm)) :
-                species
-                        .map(_species -> GeneQuery.create(queryTerm, BioentityPropertyName.getByName(category), _species))
-                        .orElseGet(() -> GeneQuery.create(queryTerm, BioentityPropertyName.getByName(category)));
+        GeneQuery geneQuery = geneIdSearchService.getGeneQueryByRequestParams(requestParams);
```
alfonsomunozpomer: `var` may be used here.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -179,6 +138,64 @@ public String search(@RequestParam MultiValueMap<String, String> requestParams)
                         "checkboxFacetGroups", ImmutableList.of(MARKER_GENE.getTitle(), ORGANISM.getTitle())));
     }
 
+    @RequestMapping(value = "/json/gene-search/marker-genes",
+            method = RequestMethod.GET,
+            produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    public Boolean isMarkerGene(@RequestParam MultiValueMap<String, String> requestParams) {
+        GeneQuery geneQuery = geneIdSearchService.getGeneQueryByRequestParams(requestParams);
```
alfonsomunozpomer: `var` here...

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -179,6 +138,64 @@ public String search(@RequestParam MultiValueMap<String, String> requestParams)
...
+        var geneIds = geneIdSearchService.search(geneQuery);
+
+        var emptyGeneIdError = geneIdEmptyValidation(geneIds);
+        if (emptyGeneIdError.isPresent()) {
+            return false;
+        }
+
+        List<Map.Entry<String, Map<String, List<String>>>> expressedGeneIdEntries =
```
alfonsomunozpomer: and `var` here too.

app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchServiceTest.java
```diff
@@ -177,4 +186,143 @@ void ifQueryHasEmptySpeciesSearchAllSpecies() {
         verify(geneIdSearchDaoMock).searchGeneIds(searchString, "flybase_gene_id");
         verify(geneIdSearchDaoMock).searchGeneIds(searchString, "wbpsgene");
     }
+
+    @Test
+    void whenEmptyQueryFieldsGivenThenReturnsError() {
+        var requestParams = new LinkedMultiValueMap<String, String>();
+        final String emptyValue = "";
```
alfonsomunozpomer, line 193: You may want to use our `RandomDataTestUtils::generateBlankString` that generates blank strings with a combination of space characters and also empty strings.

app/src/main/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchService.java
```diff
@@ -1,25 +1,44 @@
 package uk.ac.ebi.atlas.search.geneids;
 
 import com.google.common.collect.ImmutableSet;
+import org.eclipse.jetty.util.StringUtil;
```
alfonsomunozpomer: This is probably IDEA playing tricks on you, let’s stick with `org.apache.commons.lang3.StringUtils`.

app/src/test/java/uk/ac/ebi/atlas/search/JsonGeneSearchControllerWIT.java
```diff
@@ -86,7 +87,9 @@ void cleanDatabaseTables() {
 
     @BeforeEach
     void setUp() {
-        mockMvc = MockMvcBuilders.webAppContextSetup(this.wac).build();
+        mockMvc = MockMvcBuilders.webAppContextSetup(this.wac)
+                .alwaysDo(MockMvcResultHandlers.print())
```
alfonsomunozpomer, line 103: I didn’t know about this. Very helpful! :1st_place_medal:

app/src/test/java/uk/ac/ebi/atlas/search/JsonGeneSearchControllerWIT.java
```diff
@@ -189,4 +192,34 @@ void speciesParamCanAppearBeforeGeneQuery() throws Exception {
...
+        final String expectedMessage = "{\"error\":\"Error parsing query\"}\n";
+        this.mockMvc.perform(get("/json/gene-search/marker-genes").param("q", emptyGeneSearchParams))
+                .andExpect(status().isBadRequest())
+                .andExpect(content().contentType(MediaType.APPLICATION_JSON_UTF8))
+                .andExpect(content().string(expectedMessage));
+    }
+
+    @Test
+    void whenGeneIsAMarkerGeneSearchForItReturnsTrue() throws Exception {
+        var shouldBeMarkerGene = "AT2G23910";
```
alfonsomunozpomer: Rather than this, have a look at `JdbcUtils::fetchRandomMarkerGeneFromSingleCellExperiment`.

app/src/test/java/uk/ac/ebi/atlas/search/JsonGeneSearchControllerWIT.java
```diff
@@ -189,4 +192,34 @@ void speciesParamCanAppearBeforeGeneQuery() throws Exception {
...
     }
+
+    @Test
+    void whenSearchForAMarkerGeneWithEmptyValueReturnsFalse() throws Exception {
+        final String emptyGeneSearchParams = "";
+        final String expectedMessage = "{\"error\":\"Error parsing query\"}\n";
+        this.mockMvc.perform(get("/json/gene-search/marker-genes").param("q", emptyGeneSearchParams))
+                .andExpect(status().isBadRequest())
+                .andExpect(content().contentType(MediaType.APPLICATION_JSON_UTF8))
+                .andExpect(content().string(expectedMessage));
```
alfonsomunozpomer, line 215: I think it’d be more correct to compare the contents using [`json()`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/test/web/servlet/result/ContentResultMatchers.html#json-java.lang.String-). Also in L213 and 223, even though the “similarity” doesn’t have any advantage since the results are just `true` and `false`, I think you’ll get JSON-parsing for free.

app/src/test/java/uk/ac/ebi/atlas/search/JsonGeneSearchControllerWIT.java
```diff
@@ -189,4 +192,34 @@ void speciesParamCanAppearBeforeGeneQuery() throws Exception {
...
+
+        this.mockMvc.perform(get("/json/gene-search/marker-genes").param("ensgene", shouldBeMarkerGene))
+                .andExpect(status().isOk())
+                .andExpect(content().contentType(MediaType.APPLICATION_JSON_UTF8))
+                .andExpect(content().string("true"));
+    }
+
+    @Test
+    void whenGeneIsNotAMarkerGeneSearchForItReturnsFalse() throws Exception {
+        var notAMarkerGene = "NOTMarkerGene";
```
alfonsomunozpomer: Use our amazing `RandomDataTestUtils::generateRandomEnsemblGeneId` and throw dice! :game_die: There is an astronomically low probability of generating a real ID which is, additionally, a marker gene, in which case this test will fail.

app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchServiceTest.java
```diff
@@ -177,4 +186,143 @@ void ifQueryHasEmptySpeciesSearchAllSpecies() {
...
+        var generalCategory = "q";
```
alfonsomunozpomer, line 194: I’m thinking that we can have package-access for `VALID_QUERY_FIELDS`, and pick a random category from the set, and do a similar thing in the remaining tests.

alfonsomunozpomer (changes requested):
There are still a few `String` and `GeneQuery` variables (maybe others) that can be declared with `var`.

Mostly observations with how to use random data with tests. Have a look at the comments and I think this will be done and dusted.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: Done

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: Done

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: Done

app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchServiceTest.java
ke4: TBH I would rather use the simple and much cleaner `""` than some method call that might produce a similar result, but we would have to check it. I am going to leave it like this.

app/src/main/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchService.java
ke4: Good spot. Thanks. Done.

app/src/test/java/uk/ac/ebi/atlas/search/JsonGeneSearchControllerWIT.java
ke4: You are welcome. 😅

app/src/test/java/uk/ac/ebi/atlas/search/JsonGeneSearchControllerWIT.java
ke4: I did not know that the codebase has these utility methods. Thanks.
Done

app/src/test/java/uk/ac/ebi/atlas/search/JsonGeneSearchControllerWIT.java
ke4: Thanks. Done.

app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchServiceTest.java
ke4: `q` is also part of the `VALID_QUERY_FIELDS` list, so that would not fit these tests.
I would leave these tests as they are now.

app/src/test/java/uk/ac/ebi/atlas/search/JsonGeneSearchControllerWIT.java
ke4: If I replace it to `json()` then I got a `java.lang.NoClassDefFoundError` error and the test fails.
I would leave it like this.

app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchServiceTest.java
alfonsomunozpomer: Yes. My observation was more that any single field (or more than one) with only empty values, will (correctly) throw an exception. There’s nothing particular about `q` or `symbol` in this test and the next one in that regard. So I was thinking on picking a random field here, and picking two random fields in the next one.

ke4: Ok, because you wrote `remaining tests` previously. So, you only would like me to use it in those tests that throw errors because of only empty values. I can do that after lunch ;-)

alfonsomunozpomer (approved):
I think this is ready to go.

app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchServiceTest.java
ke4: Anyway, as you accepted the PR, I merged it and do something with this in my next PR.

PR #279, opened 2022-08-05
Remove hard-coded params from json cell plot controller

haideriqbal:
This story is part of the changes we are implementing as a result of Ann data. The issue here is that we use to have hardcoded plot methods which were either `umap` or `tsne`. This is not going to be the case anymore so we decided to read the plot method and plot options from the database instead of hard coding them. Further details can be found in this [pivotal story](https://www.pivotaltracker.com/story/show/181984899).


app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/JsonCellPlotGenericController.java
```diff
@@ -0,0 +1,33 @@
...
+
+import static uk.ac.ebi.atlas.utils.GsonProvider.GSON;
+
+@RestController
+@RequestMapping(value="/json/cell-plots/generic/{experimentAccession}", method= RequestMethod.GET)
+public class JsonCellPlotGenericController {
+    private final CellPlotGenericDao cellPlotGenericDao;
+
+    public JsonCellPlotGenericController(CellPlotGenericDao cellPlotGenericDao) {
+        this.cellPlotGenericDao = cellPlotGenericDao;
```
upendrakumbham, line 19: Is there any reason to skip the service layer and inject DAO class straight into the controller?

of course, I would like to know more about this concept of Generic implementation during our standup :).

app/src/test/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotGenericDaoIT.java
```diff
@@ -0,0 +1,73 @@
...
+                new ClassPathResource("fixtures/scxa_cell_group-delete.sql"),
+                new ClassPathResource("fixtures/scxa_coords-delete.sql"),
+                new ClassPathResource("fixtures/scxa_analytics-delete.sql"),
+                new ClassPathResource("fixtures/experiment-delete.sql"),
+                new ClassPathResource("fixtures/scxa_dimension_reduction-delete.sql"));
+        populator.execute(dataSource);
+    }
+
+    @Test
+    void resultIsEmpty() {
```
upendrakumbham: Please add one positive test case here to test functionality.

app/src/test/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotGenericDaoIT.java
```diff
@@ -0,0 +1,72 @@
...
+    @Test
+    void returnValidPlotOptionForValidPlotMethod() {
+        assertThat(subject.getQueryParams("UMAP", "E-CURD-4"))
+                .isNotEmpty()
+                .doesNotHaveDuplicates();
+
+    }
+
+    @Test
+    void returnEmptyListForInValidPlotMethod() {
```
upendrakumbham, line 65: Ignore my previous review comment, you have added a positive test case.

app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/JsonCellPlotGenericController.java
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    @GetMapping(value = "/params/{plotType}",
+            produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    public String getPlotParams(@PathVariable String plotType,
+                                @PathVariable String experimentAccession) {
+        var result = cellPlotGenericDao.getQueryParams(plotType, experimentAccession);
+
+        if (result.isEmpty()) {
+            throw new IllegalArgumentException("Unknown plot type " + plotType);
```
upendrakumbham, line 29: I would prefer to do this validation well before, not in the controller as we knew the result at the DAO level. Do you have a reason to do this in the controller?

app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/JsonCellPlotGenericController.java
haideriqbal: This was discussed in the stand up and @upendrakumbham agreed on the solution implemented here.

app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/JsonCellPlotGenericController.java
haideriqbal: I don't have any strong opinion on where to catch this exception. We are also doing a similar kind of exception catching in `JsonCellPlotController`. My reason for catching this exception here instead of DAO class was IMO it makes it easier to read from the controller class to see if the result is empty then it's not the right plot type.

If you still feel its better to have this in DAO class then I can move this exception there.

upendrakumbham: I agree, please merge away as I'm approving.

upendrakumbham (dismissed):
All looks good to me.

upendrakumbham (dismissed):
Good to go.

PR #280, opened 2022-08-22
Fix query for getting experiment accessions for marker gene IDs

ke4:
The query for getting experiment accessions for marker gene IDs was missing the join criteria between the `scxa_cell_group` and `scxa_cell_group_marker_genes` tables. Without that missing criteria, the query returns always all of the experiment accessions as the gene ID filter only filters the `scxa_cell_group_marker_genes` table, but does not connect that table to the other part of the SQL query.


alfonsomunozpomer (approved):
All good. Glad to get the old test back...

PR #283, opened 2022-08-26
Enable Jenkins testing in GKE cluster

alfonsomunozpomer:
Besides a few tweaks to the Jenkins file and the pod template YAML file, there were a few tests that had to be changed. The reason is that in the new SolrCloud server in the GKE cluster, i.e. `scxa-solrcloud-fast-hs`, we’re only loading the species covered in our test datasets:
- Homo sapiens
- Mus musculus
- Arabidopsis thaliana
- Oryza sativa
- Glycine max
- Solanum Lycopersicum

### DIFF
--- a/app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchDaoIT.java
+++ b/app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchDaoIT.java
@@ -89,11 +89,11 @@ class GeneIdSearchDaoIT {
     void ifNoDocumentsAreFoundInTheIntersectionReturnEmptySet() {
         var speciesNotCovered = "Oryza_sativa";
 
-        var queryBuilder = new SolrQueryBuilder<BioentitiesCollectionProxy>();
-        queryBuilder
-                .addQueryFieldByTerm(SPECIES, speciesNotCovered)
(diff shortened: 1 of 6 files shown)


app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchDaoIT.java
```diff
@@ -102,7 +103,7 @@ void ifNoDocumentsAreFoundInTheIntersectionReturnEmptySet() {
                             .findAny()
                             .map(tuples -> tuples.getString(BIOENTITY_IDENTIFIER.name()))
                             .orElseThrow(RuntimeException::new);
-
+            System.out.println("[[[ Searching for rogue gene ID " + geneId + " ]]]");
```
ke4: We probably don't need this professional debugging line, right? 😜

ke4 (changes requested):
I requested just a tiny change but otherwise looks fine. 👍

app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchDaoIT.java
alfonsomunozpomer: Right!

ke4 (approved):
LGTM 👍

PR #284, opened 2022-08-26
[Feature/182139671] Search species by gene id

ke4:
Changes:
- Endpoint: `/json/search/species`
- Associated pull request in 'atlas-web-core': [Add species_dv field to BioentitiesCollectionProxy](https://github.com/ebi-gene-expression-group/atlas-web-core/pull/112)

### DIFF
--- a/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
@@ -17,12 +17,11 @@ import uk.ac.ebi.atlas.experimentpage.ExperimentAttributesService;
 import uk.ac.ebi.atlas.model.experiment.singlecell.SingleCellBaselineExperiment;
 import uk.ac.ebi.atlas.search.geneids.GeneIdSearchService;
 import uk.ac.ebi.atlas.search.geneids.GeneQuery;
-import uk.ac.ebi.atlas.search.species.SpeciesSearchDao;
+import uk.ac.ebi.atlas.search.species.SpeciesSearchService;
 import uk.ac.ebi.atlas.trader.ExperimentTrader;
 import uk.ac.ebi.atlas.utils.StringUtil;
 
 import java.util.AbstractMap.SimpleEntry;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -49,13 +48,14 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
                                     .map(propertyName -> propertyName.name)
                                     .collect(toImmutableSet()))
                     .build();
+    public static final String ALL_CATEGORIES = "*";
 
     private final GeneIdSearchService geneIdSearchService;
     private final GeneSearchService geneSearchService;
     private final ExperimentTrader experimentTrader;
     private final ExperimentAttributesService experimentAttributesService;
 
-    private final SpeciesSearchDao speciesSearchDao;
+    private final SpeciesSearchService speciesSearchService;
 
     @RequestMapping(value = "/json/search",
                     method = RequestMethod.GET,
@@ -168,17 +168,18 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
             method = RequestMethod.GET,
             produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
     public Set<String> getSpeciesByGeneId(@RequestParam MultiValueMap<String, String> requestParams) {
-        // TODO validate input params
-//        if (isRequestParamsEmpty(requestParams)) {
-//            return false;
-//        }
+        if (isRequestParamsEmpty(requestParams)) {
+            return Set.of();
+        }
 
-        // TODO get the search test and category from the request params
-        var searchText = "";
-        var category = "";
+        var category = geneIdSearchService.getCategoryFromRequestParams(requestParams);
+        var searchText = requestParams.getFirst(category);
 
-        // TODO add a service and inject the DAO into it
-        var species = speciesSearchDao.searchSpecies(searchText, category);
+        if (category.equals("q")) {
+            category = ALL_CATEGORIES;
+        }
+
+        var species = speciesSearchService.search(searchText, category);
 
         return species.orElse(ImmutableSet.of());
     }
--- a/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
@@ -7,8 +7,7 @@ import lombok.RequiredArgsConstructor;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.http.MediaType;
 import org.springframework.util.MultiValueMap;
-import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.RequestMethod;
+import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.RequestParam;
 import org.springframework.web.bind.annotation.RestController;
 import org.springframework.web.servlet.support.ServletUriComponentsBuilder;
@@ -24,11 +23,9 @@ import java.util.AbstractMap.SimpleEntry;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.Set;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.collect.ImmutableList.toImmutableList;
-import static java.util.stream.Collectors.toList;
 import static uk.ac.ebi.atlas.search.FacetGroupName.MARKER_GENE;
 import static uk.ac.ebi.atlas.search.FacetGroupName.ORGANISM;
 import static uk.ac.ebi.atlas.utils.GsonProvider.GSON;
@@ -36,9 +33,6 @@ import static uk.ac.ebi.atlas.utils.GsonProvider.GSON;
 @RestController
 @RequiredArgsConstructor
 public class JsonGeneSearchController extends JsonExceptionHandlingController {
-    public static final String ALL_CATEGORIES = "*";
-    public static final String GENERIC_CATEGORY = "q";
-
     private final GeneIdSearchService geneIdSearchService;
     private final GeneSearchService geneSearchService;
     private final ExperimentTrader experimentTrader;
@@ -46,9 +40,7 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
 
     private final SpeciesSearchService speciesSearchService;
 
-    @RequestMapping(value = "/json/search",
-                    method = RequestMethod.GET,
-                    produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    @GetMapping(value = "/json/search", produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
     public String search(@RequestParam MultiValueMap<String, String> requestParams) {
         var geneQuery = geneIdSearchService.getGeneQueryByRequestParams(requestParams);
 
@@ -124,9 +116,7 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
                         "checkboxFacetGroups", ImmutableList.of(MARKER_GENE.getTitle(), ORGANISM.getTitle())));
     }
 
-    @RequestMapping(value = "/json/gene-search/marker-genes",
-            method = RequestMethod.GET,
-            produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    @GetMapping(value = "/json/gene-search/marker-genes", produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
     public Boolean isMarkerGene(@RequestParam MultiValueMap<String, String> requestParams) {
         var geneQuery = geneIdSearchService.getGeneQueryByRequestParams(requestParams);
         var geneIds = geneIdSearchService.search(geneQuery);
@@ -148,24 +138,18 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
         return markerGeneFacets != null && markerGeneFacets.size() > 0;
     }
 
-    @RequestMapping(value = "/json/gene-search/species",
-            method = RequestMethod.GET,
-            produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
-    public Set<String> getSpeciesByGeneId(@RequestParam MultiValueMap<String, String> requestParams) {
+    @GetMapping(value = "/json/gene-search/species", produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    public ImmutableSet<String> getSpeciesByGeneId(@RequestParam MultiValueMap<String, String> requestParams) {
         var category = geneIdSearchService.getCategoryFromRequestParams(requestParams);
         var queryTerm =
                 geneIdSearchService.getFirstNotBlankQueryField(requestParams.get(category)).orElseThrow();
-
-        if (category.equals(GENERIC_CATEGORY)) {
-            category = ALL_CATEGORIES;
-        }
 
         var species = speciesSearchService.search(queryTerm, category);
 
         return species.orElse(ImmutableSet.of());
     }
 
-    private List<Map.Entry<String, Map<String, List<String>>>> getMarkerGeneProfileByGeneIds(Optional<ImmutableSet<String>> geneIds) {
+    private ImmutableList<Map.Entry<String, Map<String, List<String>>>> getMarkerGeneProfileByGeneIds(Optional<ImmutableSet<String>> geneIds) {
         // We found expressed gene IDs, let’s get to it now...
         var geneIds2ExperimentAndCellIds =
                 geneSearchService.getCellIdsInExperiments(
@@ -173,7 +157,7 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
 
         return geneIds2ExperimentAndCellIds.entrySet().stream()
                         .filter(entry -> !entry.getValue().isEmpty())
-                        .collect(toList());
+                        .collect(toImmutableList());
     }
 
     private Optional<String> geneIdEmptyValidation(Optional<ImmutableSet<String>> geneIds) {
--- a/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
@@ -44,18 +44,15 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
                     .build();
 
     private final GeneIdSearchService geneIdSearchService;
-    private final SpeciesFactory speciesFactory;
     private final GeneSearchService geneSearchService;
     private final ExperimentTrader experimentTrader;
     private final ExperimentAttributesService experimentAttributesService;
 
     public JsonGeneSearchController(GeneIdSearchService geneIdSearchService,
-                                    SpeciesFactory speciesFactory,
                                     GeneSearchService geneSearchService,
                                     ExperimentTrader experimentTrader,
                                     ExperimentAttributesService experimentAttributesService) {
         this.geneIdSearchService = geneIdSearchService;
-        this.speciesFactory = speciesFactory;
         this.geneSearchService = geneSearchService;
         this.experimentTrader = experimentTrader;
         this.experimentAttributesService = experimentAttributesService;
--- a/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
@@ -15,6 +15,7 @@ import uk.ac.ebi.atlas.controllers.JsonExceptionHandlingController;
 import uk.ac.ebi.atlas.experimentpage.ExperimentAttributesService;
 import uk.ac.ebi.atlas.model.experiment.singlecell.SingleCellBaselineExperiment;
 import uk.ac.ebi.atlas.search.geneids.GeneIdSearchService;
+import uk.ac.ebi.atlas.search.geneids.QueryParsingException;
 import uk.ac.ebi.atlas.search.species.SpeciesSearchService;
 import uk.ac.ebi.atlas.trader.ExperimentTrader;
 import uk.ac.ebi.atlas.utils.StringUtil;
@@ -142,7 +143,9 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
     public ImmutableSet<String> getSpeciesByGeneId(@RequestParam MultiValueMap<String, String> requestParams) {
         var category = geneIdSearchService.getCategoryFromRequestParams(requestParams);
         var queryTerm =
-                geneIdSearchService.getFirstNotBlankQueryField(requestParams.get(category)).orElseThrow();
+                geneIdSearchService.getFirstNotBlankQueryField(requestParams.get(category))
+                        .orElseThrow(() -> new QueryParsingException(
+                                String.format("All fields are blank for category: %s", category)));
 
         var species = speciesSearchService.search(queryTerm, category);
 
--- a/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
@@ -16,14 +16,12 @@ import uk.ac.ebi.atlas.experimentpage.ExperimentAttributesService;
 import uk.ac.ebi.atlas.model.experiment.singlecell.SingleCellBaselineExperiment;
 import uk.ac.ebi.atlas.search.geneids.GeneIdSearchService;
 import uk.ac.ebi.atlas.search.geneids.GeneQuery;
-import uk.ac.ebi.atlas.solr.bioentities.BioentityPropertyName;
 import uk.ac.ebi.atlas.species.SpeciesFactory;
 import uk.ac.ebi.atlas.trader.ExperimentTrader;
 import uk.ac.ebi.atlas.utils.StringUtil;
 
 import java.util.*;
 import java.util.AbstractMap.SimpleEntry;
-import java.util.stream.Stream;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.collect.ImmutableList.toImmutableList;
@@ -67,54 +65,18 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
                     method = RequestMethod.GET,
                     produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
     public String search(@RequestParam MultiValueMap<String, String> requestParams) {
-        var species = Stream.ofNullable(requestParams.getFirst("species"))
-                .filter(org.apache.commons.lang3.StringUtils::isNotEmpty)
-                .map(speciesFactory::create)
-                .findFirst();
-
-        // We support currently only one query term; in the unlikely case that somebody fabricates a URL with more than
-        // one we’ll build the query with the first match. Remember that in order to support multiple terms we’ll
-        // likely need to change GeneQuery and use internally a SemanticQuery
-        var category =
-                requestParams.keySet().stream()
-                        // We rely on "q" and BioentityPropertyName::name’s being lower case
-                        .filter(actualField -> VALID_QUERY_FIELDS.contains(actualField.toLowerCase()))
-                        .findFirst()
-                        .orElseThrow(() -> new RuntimeException("Error parsing query"));
-        var queryTerm = requestParams.getFirst(category);
-
-        var geneQuery = category.equals("q") ?
-                species
-                        .map(_species -> GeneQuery.create(queryTerm, _species))
-                        .orElseGet(() -> GeneQuery.create(queryTerm)) :
-                species
-                        .map(_species -> GeneQuery.create(queryTerm, BioentityPropertyName.getByName(category), _species))
-                        .orElseGet(() -> GeneQuery.create(queryTerm, BioentityPropertyName.getByName(category)));
+        GeneQuery geneQuery = geneIdSearchService.getGeneQueryByRequestParams(requestParams);
 
         var geneIds = geneIdSearchService.search(geneQuery);
 
-        if (geneIds.isEmpty()) {
-            return GSON.toJson(
-                    ImmutableMap.of(
-                            "results", ImmutableList.of(),
-                            "reason", "Gene unknown"));
-        }
-
-        if (geneIds.get().isEmpty()) {
-            return GSON.toJson(
-                    ImmutableMap.of(
-                            "results", ImmutableList.of(),
-                            "reason", "No expression found"));
+        String emptyGeneIdError = geneIdEmptyValidation(geneIds);
+        if (emptyGeneIdError != null) {
+            return emptyGeneIdError;
         }
 
         // We found expressed gene IDs, let’s get to it now...
-        var geneIds2ExperimentAndCellIds =
-                geneSearchService.getCellIdsInExperiments(geneIds.get().toArray(new String[0]));
-
-        var expressedGeneIdEntries =
-                geneIds2ExperimentAndCellIds.entrySet().stream()
-                        .filter(entry -> !entry.getValue().isEmpty())
-                        .collect(toList());
+        List<Map.Entry<String, Map<String, List<String>>>> expressedGeneIdEntries =
+                getMarkerGeneProfileByGeneIds(geneIds);
 
         var markerGeneFacets =
                 geneSearchService.getMarkerGeneProfile(
@@ -188,20 +150,14 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
 
         GeneQuery geneQuery = geneIdSearchService.getGeneQueryByRequestParams(requestParams);
         var geneIds = geneIdSearchService.search(geneQuery);
-        String emptyGeneIdError = geneIdEmptyValidation(geneIds);
 
+        String emptyGeneIdError = geneIdEmptyValidation(geneIds);
         if (emptyGeneIdError != null) {
             return false;
         }
 
-        // We found expressed gene IDs, let’s get to it now...
-        var geneIds2ExperimentAndCellIds =
-                geneSearchService.getCellIdsInExperiments(geneIds.get().toArray(new String[0]));
-
-        var expressedGeneIdEntries =
-                geneIds2ExperimentAndCellIds.entrySet().stream()
-                        .filter(entry -> !entry.getValue().isEmpty())
-                        .collect(toList());
+        List<Map.Entry<String, Map<String, List<String>>>> expressedGeneIdEntries =
+                getMarkerGeneProfileByGeneIds(geneIds);
 
         var markerGeneFacets =
                 geneSearchService.getMarkerGeneProfile(
@@ -212,8 +168,20 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
         return markerGeneFacets != null && markerGeneFacets.size() > 0;
     }
 
+    private List<Map.Entry<String, Map<String, List<String>>>> getMarkerGeneProfileByGeneIds(Optional<ImmutableSet<String>> geneIds) {
+        // We found expressed gene IDs, let’s get to it now...
+        var geneIds2ExperimentAndCellIds =
+                geneSearchService.getCellIdsInExperiments(geneIds.get().toArray(new String[0]));
+
+        return geneIds2ExperimentAndCellIds.entrySet().stream()
+                        .filter(entry -> !entry.getValue().isEmpty())
+                        .collect(toList());
+    }
+
     private boolean isRequestParamsEmpty(MultiValueMap<String, String> requestParams) {
-        return requestParams.containsKey("q") && Objects.equals(requestParams.getFirst("q"), "");
+        return requestParams == null
+                || requestParams.size() == 0
+                || (requestParams.containsKey("q") && Objects.equals(requestParams.getFirst("q"), ""));
     }
 
     private String geneIdEmptyValidation(Optional<ImmutableSet<String>> geneIds) {
--- a/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
@@ -21,9 +21,8 @@ import uk.ac.ebi.atlas.species.SpeciesFactory;
 import uk.ac.ebi.atlas.trader.ExperimentTrader;
 import uk.ac.ebi.atlas.utils.StringUtil;
 
+import java.util.*;
 import java.util.AbstractMap.SimpleEntry;
-import java.util.List;
-import java.util.Map;
 import java.util.stream.Stream;
 
 import static com.google.common.base.Strings.isNullOrEmpty;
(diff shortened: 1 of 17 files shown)


app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
```diff
@@ -0,0 +1,122 @@
...
+        //  on="bioentity_identifier"
+        // )
+
+        if (StringUtils.isBlank(searchText)) {
+            return Optional.of(ImmutableSet.of());
+        }
+
+        UniqueStreamBuilder uniqueBioEntitiesStreamBuilder = buildUniqueStreamBuilder(searchText, category);
+
+        UniqueStreamBuilder uniqueGene2ExperimentStreamBuilder = buildUniqueStreamBuilder();
```
alfonsomunozpomer: You can use `var` here, I guess.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
```diff
@@ -0,0 +1,122 @@
...
+
+    private final BioentitiesCollectionProxy bioentitiesCollectionProxy;
+    private final Gene2ExperimentCollectionProxy gene2ExperimentCollectionProxy;
+
+    public SpeciesSearchDao(SolrCloudCollectionProxyFactory collectionProxyFactory) {
+        bioentitiesCollectionProxy = collectionProxyFactory.create(BioentitiesCollectionProxy.class);
+        gene2ExperimentCollectionProxy = collectionProxyFactory.create(Gene2ExperimentCollectionProxy.class);
+    }
+
+    public Optional<ImmutableSet<String>> searchSpecies(String searchText, String category) {
```
alfonsomunozpomer: What’s the difference in terms of semantics of this method returning `Optional.empty()` vs `Optional.of(ImmutableSet.of())`?

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -162,22 +148,34 @@ public Boolean isMarkerGene(@RequestParam MultiValueMap<String, String> requestP
         return markerGeneFacets != null && markerGeneFacets.size() > 0;
     }
 
+    @RequestMapping(value = "/json/gene-search/species",
+            method = RequestMethod.GET,
+            produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    public Set<String> getSpeciesByGeneId(@RequestParam MultiValueMap<String, String> requestParams) {
```
alfonsomunozpomer: I think it’s better practice to [declare the return type to be as specific as possible](https://en.wikipedia.org/wiki/Robustness_principle) (`ImmutableSet<String>` in this case). It gives you the same composability with less room for exceptions (e.g. let’s say that you try to add elements to the return value of this method elsewhere). However, now that I think of it, it surprises me that Spring auto-serialised the response to JSON, is this documented behaviour? All our controllers, as far as I can remember, return `String` and we explicitly serialise with GSON.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -15,50 +16,35 @@
...
 import static com.google.common.collect.ImmutableList.toImmutableList;
-import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static java.util.stream.Collectors.toList;
 import static uk.ac.ebi.atlas.search.FacetGroupName.MARKER_GENE;
 import static uk.ac.ebi.atlas.search.FacetGroupName.ORGANISM;
-import static uk.ac.ebi.atlas.solr.cloud.collections.BioentitiesCollectionProxy.BIOENTITY_PROPERTY_NAMES;
 import static uk.ac.ebi.atlas.utils.GsonProvider.GSON;
 
 @RestController
+@RequiredArgsConstructor
```
alfonsomunozpomer, line 35: Fair  enough! :D You’ll see in our project that we try to relegate magic to Google’s autovalue and Spring. We’ve never used Lombok enough to know if it goes well with Spring MVC/Boot.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -162,22 +148,34 @@ public Boolean isMarkerGene(@RequestParam MultiValueMap<String, String> requestP
...
     private List<Map.Entry<String, Map<String, List<String>>>> getMarkerGeneProfileByGeneIds(Optional<ImmutableSet<String>> geneIds) {
         // We found expressed gene IDs, let’s get to it now...
         var geneIds2ExperimentAndCellIds =
-                geneSearchService.getCellIdsInExperiments(geneIds.get().toArray(new String[0]));
+                geneSearchService.getCellIdsInExperiments(
+                        geneIds.get().toArray(new String[0]));
 
         return geneIds2ExperimentAndCellIds.entrySet().stream()
                         .filter(entry -> !entry.getValue().isEmpty())
                         .collect(toList());
```
alfonsomunozpomer: Better to collect to an `ImmutableList` and declare the return type to match. I guess we missed this one.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchService.java
```diff
@@ -0,0 +1,32 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(SpeciesSearchService.class);
+
+    private final SpeciesSearchDao speciesSearchDao;
+
+
+    public Optional<ImmutableSet<String>> search(String searchText, String category) {
+        if (StringUtils.isBlank(searchText)) {
+            LOGGER.debug("Search text is empty that is going to result with no species.");
+
+            return Optional.of(ImmutableSet.of());
```
alfonsomunozpomer: I think that in this case you want to return `Optional.empty()`, since a valid query that yields no results would return `Optional.of(ImmutableSet.of())`.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -15,50 +16,35 @@
...
 public class JsonGeneSearchController extends JsonExceptionHandlingController {
-    private final static ImmutableSet<String> VALID_QUERY_FIELDS =
-            ImmutableSet.<String>builder()
-                    .add("q")
-                    .addAll(
-                            BIOENTITY_PROPERTY_NAMES.stream()
-                                    .map(propertyName -> propertyName.name)
-                                    .collect(toImmutableSet()))
-                    .build();
+    public static final String ALL_CATEGORIES = "*";
```
alfonsomunozpomer: This is an implementation detail. I don’t think the controller should have any responsibility of knowing Solr syntax. Better to delay this choice to the DAO class.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -162,22 +148,34 @@ public Boolean isMarkerGene(@RequestParam MultiValueMap<String, String> requestP
         return markerGeneFacets != null && markerGeneFacets.size() > 0;
     }
 
+    @RequestMapping(value = "/json/gene-search/species",
```
alfonsomunozpomer: You can use `@GetMapping` here to be a bit more concise.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
```diff
@@ -0,0 +1,122 @@
...
+        //                                          sort="bioentity_identifier asc",
+        //                                          qt="/export"
+        //          ),
+        //          over="bioentity_identifier"
+        //      ),
+        //  on="bioentity_identifier"
+        // )
+
+        if (StringUtils.isBlank(searchText)) {
+            return Optional.of(ImmutableSet.of());
```
alfonsomunozpomer: You’re checking this in the service class, too.

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDaoIT.java
```diff
@@ -0,0 +1,99 @@
...
+
+    private SpeciesSearchDao subject;
+
+    @BeforeEach
+    void setUp() {
+        subject = new SpeciesSearchDao(collectionProxyFactory);
+    }
+
+    @Test
+    void whenEmptySearchTextProvidedReturnEmptyOptional() {
```
alfonsomunozpomer, line 35: The name of this test is a bit misleading because the Optional isn’t empty. See my comments above regarding when I think this method should return an empty optional vs an optional containing an empty collection.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
```diff
@@ -0,0 +1,122 @@
...
+        var speciesStreamQueryBuilder = new InnerJoinStreamBuilder(
+                uniqueBioEntitiesStreamBuilder,
+                uniqueGene2ExperimentStreamBuilder,
+                BIOENTITY_IDENTIFIER.name()
+        );
+
+        return getSpeciesFromStreamQueryBuilder(speciesStreamQueryBuilder);
+    }
+
+    private Optional<ImmutableSet<String>> getSpeciesFromStreamQueryBuilder(InnerJoinStreamBuilder speciesStreamQueryBuilder) {
```
alfonsomunozpomer: I think `Query` just adds a bit of noise, compare with:
```java
private Optional<ImmutableSet<String>> getSpeciesFromStreamBuilder(InnerJoinStreamBuilder speciesStreamBuilder) {
```

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
```diff
@@ -0,0 +1,122 @@
...
+        var searchBioEntitiesBuilder =
+                new SearchStreamBuilder<>(bioentitiesCollectionProxy, bioEntitiesByTextAndCategory).returnAllDocs();
+        var selectBioEntitiesStreamBuilder =
+                new SelectStreamBuilder(searchBioEntitiesBuilder)
+                        .addFieldMapping(ImmutableMap.of(
+                                BIOENTITY_IDENTIFIER_DV.name(), BIOENTITY_IDENTIFIER.name(),
+                                SPECIES_DV.name(), SPECIES.name()
+                        ));
+        return new UniqueStreamBuilder(selectBioEntitiesStreamBuilder, BIOENTITY_IDENTIFIER.name());
+    }
```
alfonsomunozpomer: I’m not a big fan of throw-away methods or variables unless they introduce clarity of concepts in the code. For instance, `getSpeciesFromStreamBuilder` encapsulates under a name what it does, but `buildUniqueStreamBuilder` not so much. In the end is a matter of taste and I’ll leave it to your judgment. Just some food for thought.

alfonsomunozpomer (changes requested):
Have a look at the detailed comments I left. I think that the the most important bit is the return value of the `Optional` is something that can be iterated upon (as it is, I think it will never return an empty optional).

P.S.: I see that eventually `SearchTestUtil` made it to a reused class in more tests.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
ke4: Done

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
ke4: The 1st one returns an empty `Optional`.
The 2nd one returns an `Optional` that contains an empty `ImmutableSet`. A `Set` that has no elements.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: I changed it to `ImmutableSet<String>`.
When I tested it with Postman I got a proper JSON response from this endpoint. I am going to check it again when I went through all of the review comments.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
alfonsomunozpomer: I was asking from a semantics point of view (i.e. what is it telling you in each case, such as not found or the query not being valid, etc.).

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: I am always trying to use Lombok as it saves me writing a lot of boilerplate codes. You don't have to use everything from Lombok. `@RequiredArgsConstructor` is just a very simple replacement of a constructor initializing all its required fields. The code looks much clearer using some of these Lombok features. They are NOT magic 😅

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchService.java
ke4: Good idea. I change it, but we probably never get a blank search term here as we already check this previously and throw an exception. I checked this behaviour also with Postman.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: Good spot. I moved it there and changed the code and a test that was badly written.
Also, it seems that sending `*` as `ALL_CATEGORY` is not working well with our current SolR library as the parsing code change `"q"` to `"\q"` (escaping it as a special character) and that won't give us a correct result when we use that value is our streaming SolR expression.
We can achieve the correct (same) result if we exclude this value from the streaming expression if the value is `q`. It takes less time than debugging the SolR library.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: That is true. I changed it at the other endpoints, too.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
ke4: Yes, but it is a public method, so anybody could call this with a blank `searchTerm`.
I would leave this validation check here.

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDaoIT.java
ke4: As I changed the code in the class under test now the test method name makes sense.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
ke4: Agreed. Renamed.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
ke4: Yep. I also realized it after I pushed it to GH, but I thought I am going to rename these badly named methods when I do the change requests of the code review or in my next PR.
I think I used IntelliJ's refactoring tool and I for some insane reason forgot to rename the method names provided by IntelliJ. Sorry about that.

ke4: Renamed a bunch of methods and variables. Hopefully, these names are more meaningful.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: Yep. All is fine. We get a proper JSON response.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
ke4: OK, I got you now. Sorry.
The 1st could mean that we have some issues processing the request, because if we could process the request successfully but we got an empty result set, that's when we response with `Optional.of(ImmutableSet.of())`.
Is that what you meant?

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: I tested this new implementation and it works fine.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: Done

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -162,20 +138,26 @@ public Boolean isMarkerGene(@RequestParam MultiValueMap<String, String> requestP
         return markerGeneFacets != null && markerGeneFacets.size() > 0;
     }
 
-    private List<Map.Entry<String, Map<String, List<String>>>> getMarkerGeneProfileByGeneIds(Optional<ImmutableSet<String>> geneIds) {
+    @GetMapping(value = "/json/gene-search/species", produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    public ImmutableSet<String> getSpeciesByGeneId(@RequestParam MultiValueMap<String, String> requestParams) {
+        var category = geneIdSearchService.getCategoryFromRequestParams(requestParams);
+        var queryTerm =
+                geneIdSearchService.getFirstNotBlankQueryField(requestParams.get(category)).orElseThrow();
```
alfonsomunozpomer: How about throwing a new `QueryParsingException` with a human friendly message? I just remembered about this class from your earlier PR.

app/src/main/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchService.java
```diff
@@ -129,7 +129,7 @@ public String getCategoryFromRequestParams(MultiValueMap<String, String> request
         throw new QueryParsingException("Error parsing query");
     }
 
-    private Optional<String> getFirstNotBlankQueryField(List<String> values) {
+    public Optional<String> getFirstNotBlankQueryField(List<String> values) {
         for (var searchValue : values) {
             if (StringUtils.isNotBlank(searchValue)) {
                 return Optional.of(searchValue);
```
alfonsomunozpomer: Not strictly part of this PR, but I realised that you can express this more succinctly with:
```java
return values.stream().filter(StringUtils::isNotBlank).findFirst();
```

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
```diff
@@ -0,0 +1,135 @@
...
+
+        var streamBuilderForUniqueSpeciesFromBioEntities =
+                getStreamBuilderForUniqueSpeciesFromBioEntities(searchTerm, category);
+
+        var streamBuilderForUniqueBioEntityIDFromGene2Experiment =
+                getStreamBuilderForUniqueBioEntityIDFromGene2Experiment();
+
+        var streamBuilderForSpecies = new InnerJoinStreamBuilder(
+                streamBuilderForUniqueSpeciesFromBioEntities,
+                streamBuilderForUniqueBioEntityIDFromGene2Experiment,
```
alfonsomunozpomer: I think it’s cleaner if you inline the private method calls here rather than using variables: fewer degrees of indirection.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
```diff
@@ -0,0 +1,135 @@
...
+        //  on="bioentity_identifier"
+        // )
+
+        if (StringUtils.isBlank(searchTerm)) {
+            LOGGER.info("We can't conduct a search with an empty search term.");
+            return Optional.empty();
+        }
+
+        if (category != null && category.equals(GENERIC_CATEGORY)) {
+            LOGGER.info("We can't use generic category (q) as a filter, so we set it to null.");
```
alfonsomunozpomer: Rather than a log message, I think this is better left as a comment in the code.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
```diff
@@ -0,0 +1,135 @@
...
+            category = null;
```
alfonsomunozpomer: Instead of setting `category` to null, why don’t we leave this decision to `getStreamBuilderForUniqueSpeciesFromBioEntities`. See my other comment below.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
```diff
@@ -0,0 +1,135 @@
...
+        return new UniqueStreamBuilder(searchBuilderForBioEntities, BIOENTITY_IDENTIFIER.name());
+    }
+
+    private UniqueStreamBuilder getStreamBuilderForUniqueSpeciesFromBioEntities(String searchText, String category) {
+        var bioEntitiesByTextAndCategory = new SolrQueryBuilder<BioentitiesCollectionProxy>()
+                .addQueryFieldByTerm(PROPERTY_VALUE, searchText)
+                .setFieldList(ImmutableSet.of(SPECIES_DV, BIOENTITY_IDENTIFIER_DV))
+                .sortBy(BIOENTITY_IDENTIFIER_DV, SolrQuery.ORDER.asc);
+
+        if (StringUtils.isNotBlank(category)) {
```
alfonsomunozpomer:
```java
if (!category.equals(GENERIC_CATEGORY)) {
    ...
}
```

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchService.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private final SpeciesSearchDao speciesSearchDao;
+
+
+    public Optional<ImmutableSet<String>> search(String searchText, String category) {
+        if (StringUtils.isBlank(searchText)) {
+            LOGGER.debug("Search text is empty that is going to result with no species.");
+
+            return Optional.empty();
+        }
```
alfonsomunozpomer: The DAO class is doing this for you. However, if you remove this check then your service class isn’t doing any work at all. This means that maybe the DAO class is doing too much work? Maybe it’s the service who should decide whether to narrow by category or not, something like:
```java
if (category.equals(GENERIC_CATEGORY)) {
    return speciesSearchDao(searchSpecies(searchText));
} else {
    return speciesSearchDao(searchSpecies(searchText, category));
}
```

Or maybe leave the `isBlank` checks here, but they seem to me redundant.

app/src/test/java/uk/ac/ebi/atlas/search/JsonGeneSearchControllerIT.java
```diff
@@ -138,4 +144,66 @@ void whenGeneIsAMarkerGeneSearchForItReturnsTrue() {
...
+    @Test
+    void whenRequestParamIsNullSpeciesSearchReturnsEmptySet() {
+        LinkedMultiValueMap<String, String> requestParams = null;
+
+        when(geneIdSearchServiceMock.getCategoryFromRequestParams(requestParams))
+                .thenThrow(new QueryParsingException("Error parsing query"));
+
+        assertThatExceptionOfType(QueryParsingException.class)
+                .isThrownBy(() -> subject.getSpeciesByGeneId(requestParams));
+    }
```
alfonsomunozpomer: The name of these two tests aren’t right, you’re checking that the subject class throws an exception.

app/src/test/java/uk/ac/ebi/atlas/search/JsonGeneSearchControllerIT.java
```diff
@@ -138,4 +144,66 @@ void whenGeneIsAMarkerGeneSearchForItReturnsTrue() {
...
+        requestParams.add(generalCategory, notPartOfAnyExperiment);
+
+        when(geneIdSearchServiceMock.getCategoryFromRequestParams(requestParams))
+                .thenReturn(generalCategory);
+        when(geneIdSearchServiceMock.getFirstNotBlankQueryField(List.of(notPartOfAnyExperiment)))
+                .thenReturn(Optional.of(notPartOfAnyExperiment));
+        when(speciesSearchService.search(notPartOfAnyExperiment, generalCategory))
+                .thenReturn(Optional.of(ImmutableSet.of()));
+
+        Set<String> emptySpeciesResult = subject.getSpeciesByGeneId(requestParams);
```
alfonsomunozpomer: `var` can be used here and below, L204.

app/src/test/java/uk/ac/ebi/atlas/search/JsonGeneSearchControllerIT.java
```diff
@@ -138,4 +144,66 @@ void whenGeneIsAMarkerGeneSearchForItReturnsTrue() {
...
+
+        assertThat(emptySpeciesResult).isEmpty();
+    }
+
+    @Test
+    void whenGeneIdIsPArtOfSomeExperimentsThenReturnsSetOfSpecies() {
+        var requestParams = new LinkedMultiValueMap<String, String>();
+        final String mostInterestingGeneEver = "MostInterestingGeneEver";
```
alfonsomunozpomer: :rofl:

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDaoIT.java
alfonsomunozpomer: Sorry, my bad. I misunderstood the test. You can make it more fluent with:
```java
assertThat(species).isEmpty();
```

Or if you want to make explicit that it’s checking the `Optional` rather than the `ImmutableSet`:
```java
assertThat(species).isNotPresent();
```

You can have a look at all the `Optional` assertions here: https://www.javadoc.io/doc/org.assertj/assertj-core/latest/org/assertj/core/api/AbstractOptionalAssert.html.

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDaoIT.java
```diff
@@ -0,0 +1,99 @@
...
+    @Test
+    void whenNoCategoryButValidGeneIdProvidedThenReturnResult() {
+        var searchText = "ACRV1";
+        String category = null;
+        var expectedSpecies = "Homo_sapiens";
+
+        var species = subject.searchSpecies(searchText, category);
+
+        assertThat(species.isPresent()).isTrue();
+        assertThat(species.get()).contains(expectedSpecies);
```
alfonsomunozpomer: Again, you can conflate these two tests into:
```java
assertThat(species).contains(ImmutableSet.of(expectedSpecies));
```

Or any other variant. If you’re making an assertion over the value you don’t need to check that it’s present.

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDaoIT.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Test
+    void whenNoDocumentsAreFoundInBioEntitiesReturnEmptyOptional() {
+        var species = subject.searchSpecies("FOOBAR", SYMBOL.name);
+
+        assertThat(species.isPresent()).isTrue();
+        assertThat(species.get()).isEmpty();
```
alfonsomunozpomer: Same as above. If you prefer you can use `hasValueSatisfying(Collection foo -> foo.isEmpty())`.

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchServiceTest.java
```diff
@@ -0,0 +1,71 @@
...
+        var expectedSpecies = "Homo_sapiens";
+
+        when(speciesSearchDao.searchSpecies(existingGeneIdForHuman, searchCategory))
+                .thenReturn(Optional.of(ImmutableSet.of(expectedSpecies)));
+
+        var species = subject.search(existingGeneIdForHuman, searchCategory);
+
+        assertThat(species.isPresent()).isTrue();
+        assertThat(species.get()).hasSize(1);
+        assertThat(species.get()).contains(expectedSpecies);
```
alfonsomunozpomer: Again, the last test checks for three tests.

alfonsomunozpomer (changes requested):
I’ve noticed there’s some redundancy in the methods `getCategoryFromRequestParams` and `getFirstNotBlankQueryField`. Notice that `getCategoryFromRequestParams` uses `getFirstNotBlankQueryField` which we then again call in the next line, `var queryTerm=...`. I guess we can refactor these two methods so that it returns a `Pair<String, String>` or a singleton map. This is something we don’t need to do right now, but probably at a later stage when we put all the new endpoints that get marker genes, species, organism parts and cell types together. I think all of them could reuse it.

The other bit that caught my eye is the responsibility of the service class and the DAO class. Currently the former is very thin, it’s not bad, sometimes it happens, but it’s worth having a second look.

Other than that, have a look at the assertions but this is almost done now!

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: Done

app/src/main/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchService.java
ke4: Done

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
ke4: yep, after renaming most of these methods I forgot to simplify it further. Thanks.

ke4: Done

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
ke4: Done

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchService.java
ke4: Done. I left the `isBlank` check here.

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
ke4: Done

app/src/main/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDao.java
ke4: Removed this code segment

app/src/test/java/uk/ac/ebi/atlas/search/JsonGeneSearchControllerIT.java
ke4: Good spot. Done.

app/src/test/java/uk/ac/ebi/atlas/search/JsonGeneSearchControllerIT.java
ke4: Done

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDaoIT.java
ke4: Done

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDaoIT.java
ke4: Done

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchServiceTest.java
ke4: Done

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -162,20 +139,28 @@ public Boolean isMarkerGene(@RequestParam MultiValueMap<String, String> requestP
...
+    public ImmutableSet<String> getSpeciesByGeneId(@RequestParam MultiValueMap<String, String> requestParams) {
+        var category = geneIdSearchService.getCategoryFromRequestParams(requestParams);
+        var queryTerm =
+                geneIdSearchService.getFirstNotBlankQueryField(requestParams.get(category))
+                        .orElseThrow(() -> new QueryParsingException(
+                                String.format("All fields are blank for category: %s", category)));
+
+        var species = speciesSearchService.search(queryTerm, category);
+
+        return species.orElse(ImmutableSet.of());
```
alfonsomunozpomer: Notice that whenever your service returns an empty optional, rather than any having an alternate behaviour or a dynamic value, you’re converting it to an empty set. In that case, why do you need an optional at all? I think that you’ll get exactly the same functionality without optionals. Have the service/DAO class return `ImmutableSet.of()` when they return `Optional.empty()` and it’s a layer of complexity that can be removed.

app/src/test/java/uk/ac/ebi/atlas/experimentpage/TSnePlotSettingsServiceIT.java
```diff
@@ -122,6 +123,7 @@ void getPerplexitiesForValidAccession() {
     // dataFileHub.getSingleCellExperimentFiles(experimentAccession).clustersTsv.get().get()
     // The +1 magic number accounts for open DB connections in the build environment, in my laptop it’s +4.
     // This is the next best thing I could come up with... sorry! :(
+    @Ignore // I added this annotation for now to turn this test of as we discussed this on the stand-up meeting
```
alfonsomunozpomer: As we recently found out, you want this to be `@Disabled`. :wink:

app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchServiceTest.java
```diff
@@ -288,11 +296,14 @@ void whenRequestParamsEmptyThenThrowQueryParsingException() {
 
     @Test
     void whenRequestParamsHasGenericQueryFieldThenGotProperGeneQuery() {
```
alfonsomunozpomer: This test is testing that you get the expected query if the species is omitted... (continues below)

app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchServiceTest.java
```diff
@@ -302,14 +313,16 @@ void whenRequestParamsHasGenericQueryFieldThenGotProperGeneQuery() {
     @Test
     void whenRequestParamsHasValidQueryFieldThenGotProperGeneQuery() {
```
alfonsomunozpomer: ... and this test is testing that the species field is included in the expected gene query.

The names should be changed to reflect that.

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDaoIT.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Test
+    void whenNoCategoryButValidGeneIdProvidedThenReturnResult() {
+        var searchText = "ACRV1";
+        String category = null;
+        var expectedSpecies = ImmutableSet.of("Homo_sapiens", "Mus_musculus");
+
+        var species = subject.searchSpecies(searchText, category);
+        System.out.println(species);
```
alfonsomunozpomer: Remove enterprise-level logging framework. :smile:

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDaoIT.java
```diff
@@ -0,0 +1,97 @@
...
+    @Test
+    void whenNoDocumentsAreFoundInTheIntersectionReturnEmptySet() {
+        // Gorilla's gene exists in our Solr BioEntities collection, but not in our experiments
+        // So, the species search by gene id should not find any results
+        var gorillaGeneNotInOurExperiment = "ENSGGOG00000029327";
+
+        var species = subject.searchSpecies(gorillaGeneNotInOurExperiment, SYMBOL.name);
+
+        assertThat(species.isPresent()).isTrue();
+        assertThat(species).hasValue(ImmutableSet.of());
```
alfonsomunozpomer: You can get a less verbose version of the first assertion as:
```java
assertThat(species).isPresent();
```

However, notice that `hasValue(...)` includes the first assertion, so there’s no need to have the former. The same applies to the two tests below: if `species` is not present, the second assertion will always fail.

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchServiceTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Test
+    void whenEmptySearchTextProvidedReturnEmptyOptional() {
+        var emptySearchText = "";
+        var searchCategory = SYMBOL.name();
+
+        var emptyListOfSpecies = subject.search(emptySearchText, searchCategory);
+
+        assertThat(emptyListOfSpecies.isEmpty()).isTrue();
```
alfonsomunozpomer: This can be:
```java
assertThat(emptyListOfSpecies).isEmpty();
```

alfonsomunozpomer (changes requested):
I still have my reservations about wrapping the return type of the service and the DAO in `Optional`. In the end you’re treating it as an empty collection. Maybe, though, I can be missing something in the flow of data.

Also, there are a few things to change in the tests (especially the name of two, as you can see in the review notes). I can tell that certain habits from Hamcrest are hard to change (it’s common to have `assertThat(/*function that wrangles actual result to a boolean condition*/).isTrue()` but AssertJ has many different constructs to make all those tests more readable.

app/src/test/java/uk/ac/ebi/atlas/experimentpage/TSnePlotSettingsServiceIT.java
ke4: Yep. Done

app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchServiceTest.java
ke4: Renamed.

app/src/test/java/uk/ac/ebi/atlas/search/geneids/GeneIdSearchServiceTest.java
ke4: Renamed.

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDaoIT.java
ke4: 😅 Done.

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDaoIT.java
ke4: True for this one, but not for the 2 below.
You have to check if the `Optional` is present or not. Without that check, calling `get()` is not safe.

ke4: Fixed this one and left untouched the other 2 for now.

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchServiceTest.java
ke4: Done

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: Agreed. I just did it to follow other implementations in `GeneIdSearchDao/Service`.
I guess eventually we could clean those up as well in another PR.

app/src/test/java/uk/ac/ebi/atlas/search/species/SpeciesSearchDaoIT.java
ke4: As I removed `Optional` wrapper from the results I could clean up those 2 as well.

alfonsomunozpomer (approved):
No objections. LGTM!

PR #289, opened 2022-09-06
Remove gene search tab

### LINKED ISSUE #286: Move gene search result into an empty tab like metadata search result in SC

lingyun1010:
This story is about removing gene search tab from SC homepage, so the gene search results will show as an individual page without highlighting any tab.


alfonsomunozpomer (approved):
All good.

haideriqbal (approved):
LGTM!

PR #290, opened 2022-09-09
Fix colour by category on cell plots

### LINKED ISSUE #282: No colouring in experiments without inferred cell type ontology/authors labels
Experiments which colour plots by a k value show no clusters by default. Manually selecting the value from the drop-down renders the plot correctly.

Examples:
- https://www.ebi.ac.uk/gxa/sc/experiments/E-CURD-102/results/tsne
- https://www.ebi.ac.uk/gxa/sc/experiments/E-CURD-90/results/tsne

image
image

haideriqbal:
The issue arose when we last fixed the URL routing issue where there was no colouring. (https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/239).

The issue is when the component is initialised and there are no `cell_type` options available in the metadata, then it still selects `metadata` as `selectedColourByCategory` which is wrong. So now I have fixed it and depending on if there is a `cell_type` metadata then the category is metadata otherwise clusters which is the correct behaviour.


upendrakumbham (approved):
LGTM!

PR #299, opened 2022-10-11
- new container added to scale up the solrcloud and zookeper stateful…

davidg251:
…set before run the build

- new stage in the pipeline to execute the kubectl container


ke4:
This has been already implemented in this PR already: #323

PR #300, opened 2022-10-21
Filter out not applicable from cell type wheel

### LINKED ISSUE #258: Not applicable cell type shows in the cell type results
When searching for lung in the metadata through https://wwwdev.ebi.ac.uk/gxa/sc/search/metadata/lung

we see not applicable cell types in the wheel, which we should probably remove / filter out:

image

lingyun1010:
When searching for lung in the metadata through https://wwwdev.ebi.ac.uk/gxa/sc/search/metadata/lung

We see not applicable cell types in the wheel, which we filter it out in solr queries.


alfonsomunozpomer (changes requested):
The tests in Jenkins are not passing. Also add tests for this fix, please.

alfonsomunozpomer (approved):
Clean and simple.

PR #301, opened 2022-11-01
Add search box and species drop-down to cell type wheel

ke4:
Changes:

- Using the updated `@ebi-gene-expression-group/scxa-cell-type-wheel-heatmap` component. The search box field and the species drop-down have been added to this component.
- The height of the search box has been fixed, too.


alfonsomunozpomer (approved):
LGTM.

PR #303, opened 2022-11-23
[Feature/182154560] [Backend] New gene search endpoint

ke4:
Changes:

This change adds a new endpoint for gene search (`/json/gene-search`). This is more or less the duplicate of the existing gene search endpoint (`/json/search`) with a small change. We are not generating the data for the UI facets (checkbox and dropdown (multi-select) groups) anymore in the backend. We move it to the front end. That means we could also simplify the structure of the `results` key in the response of this endpoint removing the facets from there.
When the gene search refactors/epic has finished we can safely deprecate and remove the current gene search endpoint and probably we could do a bit more code cleaning/simplifying.


lingyun1010 (approved):
Looks good for me! But please rebase the `develop` branch firstly before merging.

PR #307, opened 2022-12-14
Make HealthCheckController check for bioentities as an alias

alfonsomunozpomer:
We’re now aliasing the bioentitites collection as `bioentities-vx` -> `bioentities`.


ke4 (approved):
@alfonsomunozpomer Would be better to read it from ENV variables, but as a quick fix, it is fine.

PR #308, opened 2022-12-15
Improve development Docker environment

alfonsomunozpomer:
Besides a rogue `sleep` statement that had made its way to the Solr loader container, I’ve added a new volume to keep the Gradle wrapper packages so we don’t have to download them every time we run tests. Just remember to re-run the `./docker/prepare-dev-environment/gradle-ro-dep-cache/run.sh` to get the latest goodies.


docker/prepare-dev-environment/solr/docker-entrypoint.sh
```diff
@@ -49,5 +49,3 @@ do
   MATRIX_MARKT_ROWS_GENES_FILE=/atlas-data/scxa/magetab/${EXP_ID}/${EXP_ID}.aggregated_filtered_normalised_counts.mtx_rows \
   ./load-scxa-gene2experiment.sh
 done
-
-sleep 3h
```
ke4, line 53: I already commented on this in your previous PR 😅

ke4 (approved):
I already asked if we need that 3h sleep in your previous PR, but never mind 😅
Looks fine to me 👍

PR #309, opened 2022-12-15
Quick-fix for local docker environment for Solr

ke4:
Changes:
- Add Dsolr.max.booleanClauses=100000000 to local Solr's docker-compose config


alfonsomunozpomer (approved):
Because I like to live dangerously, I’m approving this before getting a green build.

PR #310, opened 2022-12-15
Add species as a request parameter for metadata search endpoint

alfonsomunozpomer:
I added a request parameter to both the JSON and the view/HTML controller. The service and DAO classes that are called from the JSON endpoint have now two methods to search with/without a species. The HTML controller, in turn, passes the parameter as a prop to the React component. The React comopnent needs to be aware of this prop in a following story.


app/src/main/java/uk/ac/ebi/atlas/search/metadata/CellTypeWheelDao.java
```diff
@@ -23,6 +23,47 @@
...
+//                 }
+//               }
+//             }
+//           }
+//         }
+//       }
+//     }
+//   }
+// }
+// '
```
ke4, line 69: If we need this comment here, maybe a short 1 sentence summary would be good at the top of these lines.

app/src/main/java/uk/ac/ebi/atlas/search/metadata/CellTypeWheelDao.java
```diff
@@ -121,6 +128,23 @@ public ImmutableList<ImmutableList<String>> facetSearchCtwFields(String searchTe
...
 
+    public ImmutableList<ImmutableList<String>> facetSearchCtwFields(String searchTerm) {
+        var facetBuilder = buildFacetOrganismToOrganismPartsToCellTypesToExperimentAccessions();
+        var queryBuilder = buildQueryForFacetOfOntologyAnnotationsWithoutNotApplicableCellTypes(searchTerm)
+                .addFacet(ORGANISMS_TERMS_KEY, facetBuilder);
+
+        return extractNestedFacetTermsOrganismsAndOrganismPartsAndCellTypesAndExperimentAccessions(queryBuilder);
+    }
+
+    public ImmutableList<ImmutableList<String>> facetSearchCtwFields(String searchTerm, String species) {
```
ke4: Could we add `.addFilterFieldByTerm(CTW_ORGANISM, species)` conditionally if the `species` parameter is not `null`? If that works, then we could remove the other `facetSearchCtwFields` method.

app/src/main/java/uk/ac/ebi/atlas/search/metadata/CellTypeWheelService.java
```diff
@@ -19,7 +19,16 @@ public CellTypeWheelService(CellTypeWheelDao cellTypeWheelDao) {
     }
 
     public ImmutableSet<ImmutablePair<ImmutableList<String>, String>> search(String searchTerm) {
-        return cellTypeWheelDao.facetSearchCtwFields(searchTerm)
+        return explodeTuples(cellTypeWheelDao.facetSearchCtwFields(searchTerm));
```
ke4: If you could refactor the `facetSearchCtwFields`, then please also change this line of code. Thanks

ke4 (changes requested):
I added some really tiny comments. Could you review it, please?
Otherwise it looks fine for me 👍

alfonsomunozpomer:
@ke4 Please have another look. I reworked the controller/service/DAO classes to delay the optional behaviour all the way to the DAO class, so we don’t have overloaded methods. Initially I thought this was clearer in terms of intent, but maybe this is clearer and easier to read.

ke4 (approved):
@alfonsomunozpomer Thanks for the changes.
LGTM 👍

PR #317, opened 2023-02-06
[Bugfix] Correct the assertion in the test as it need to be case insensitive

ke4:
I had to investigate `SuggesterDaoIT` test as it looked like a flaky test in our CI system (Jenkins).
Changes:
- First I renamed it as I realised that we have a test with exactly the same class and package name in our `atlas-web-core` submodule.
- Changed the assertion in the test and use case insensitive equality checks for the properties of `Suggestion`


alfonsomunozpomer (approved):
All good, hopefully this will put an end to those pesky red builds!

PR #324, opened 2023-02-28
Experiment Design backend functionality

haideriqbal:
This PR implements relevant classes to create experiment design table in the web application by fetching data from database. The accompanying PR in [atlas-web-core](https://github.com/ebi-gene-expression-group/atlas-web-core/pull/123)


app/src/test/java/uk/ac/ebi/atlas/experiments/ExperimentDesignControllerWIT.java
```diff
@@ -0,0 +1,107 @@
...
+        populator.execute(dataSource);
+    }
+
+    @BeforeEach
+    void setUp() {
+        this.mockMvc = MockMvcBuilders.webAppContextSetup(this.wac).build();
+    }
+
+    @Test
+    void hasExperimentDesignDataWithoutAnyParameter() throws Exception {
```
ke4: This method name is not very descriptive and not telling us what is it testing.
Here is an article about how to write a descriptive test method name: https://methodpoet.com/unit-test-method-naming-convention/

app/src/test/java/uk/ac/ebi/atlas/experiments/ExperimentDesignControllerWIT.java
```diff
@@ -0,0 +1,107 @@
...
+        mockMvc.perform(get(URL, "E-EHCA-2"))
```
ke4: Could we add a bit more check on the expected response payload, please?
Could you do it for the next test, too? Thanks

app/src/main/java/uk/ac/ebi/atlas/experimentpage/tabs/ExperimentPageContentSerializer.java
```diff
@@ -35,10 +35,10 @@ public String experimentPageContentForExperiment(String experimentAccession, fin
         var results = experimentPageContentService.getTsnePlotData(experiment.getAccession());
         availableTabs.add(customContentTab("results", "Results", results));
 
-        var experimentDesignJson =
+        /*var experimentDesignJson =
```
ke4: Is this accidentally left commented out? Thanks

app/src/main/java/uk/ac/ebi/atlas/experiments/ExperimentDesignController.java
```diff
@@ -0,0 +1,33 @@
...
+    private final ExperimentDesignTableService experimentDesignService;
+
+    public ExperimentDesignController (ExperimentDesignTableService experimentDesignService) {
+        this.experimentDesignService = experimentDesignService;
+    }
+
+    @GetMapping(value = "/json/experiment-design/{experiment_accession}",
+                produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    public String getExperimentDesign(@PathVariable String experiment_accession,
+                                      @RequestParam(value = "pageNo", defaultValue = "1") int pageNo,
```
ke4: For these to magic number: `1` and `20` I suggest to create 2 constant variables with a descriptive name. Thanks

app/src/main/java/uk/ac/ebi/atlas/experiments/ExperimentDesignController.java
```diff
@@ -0,0 +1,33 @@
...
+public class ExperimentDesignController extends JsonExceptionHandlingController {
+    private final ExperimentDesignTableService experimentDesignService;
+
+    public ExperimentDesignController (ExperimentDesignTableService experimentDesignService) {
+        this.experimentDesignService = experimentDesignService;
+    }
+
+    @GetMapping(value = "/json/experiment-design/{experiment_accession}",
+                produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    public String getExperimentDesign(@PathVariable String experiment_accession,
```
ke4:
```suggestion
public String getExperimentDesign(@PathVariable String experimentAccession,
```

app/src/main/java/uk/ac/ebi/atlas/experiments/ExperimentDesignController.java
```diff
@@ -0,0 +1,33 @@
...
+                                      @RequestParam(value = "pageNo", defaultValue = "1") int pageNo,
+                                      @RequestParam(value = "pageSize", defaultValue = "20") int pageSize) {
+        if(pageNo < 1)
+            throw new IllegalArgumentException("Page number must be greater than 0");
+        if(pageSize < 1)
+            throw new IllegalArgumentException("Page size must be greater than 0");
+
+        return  GSON.toJson(experimentDesignService.getExperimentDesignData(experiment_accession, pageNo, pageSize));
```
ke4:
```suggestion
return GSON.toJson(experimentDesignService.getExperimentDesignData(experimentAccession, pageNo, pageSize));
```

ke4 (changes requested):
I added some change requests and some suggestions that you can quickly apply.
Thanks

app/src/main/java/uk/ac/ebi/atlas/experimentpage/tabs/ExperimentPageContentSerializer.java
```diff
@@ -34,11 +34,7 @@ public String experimentPageContentForExperiment(String experimentAccession, fin
 
         var results = experimentPageContentService.getTsnePlotData(experiment.getAccession());
         availableTabs.add(customContentTab("results", "Results", results));
-
-        var experimentDesignJson =
-                experimentPageContentService.getExperimentDesign(
-                        experiment.getAccession(), new ExperimentDesignTable(experiment).asJson(), accessKey);
-        availableTabs.add(customContentTab("experiment-design", "Experiment Design", experimentDesignJson));
+        availableTabs.add(customContentTab("experiment-design", "Experiment Design", new JsonObject()));
```
ke4: Is there a reason you removed the line that gets the experiment design in a JSON format from the service and you add an empty `JsonObject` instead of that? Thanks

app/src/test/java/uk/ac/ebi/atlas/experiments/ExperimentDesignControllerWIT.java
```diff
@@ -0,0 +1,113 @@
...
+    @Test
+    void hasExperimentDesignDataWithoutAnyPageNoSizeParameter() throws Exception {
+        mockMvc.perform(get(URL, "E-EHCA-2"))
+                .andExpect(status().isOk())
+                .andExpect(content().contentType(MediaType.APPLICATION_JSON_UTF8))
+                .andExpect(jsonPath("$.headers").isArray())
+                .andExpect(jsonPath("$.headers").isNotEmpty())
+                .andExpect(jsonPath("$.data").isArray())
+                .andExpect(jsonPath("$.data").isNotEmpty())
+                .andExpect(jsonPath("$.headers").contains(List.of("contrastName", "referenceOrTest", "run")))
```
ke4: I don't really like using magic strings here. Would be nice to introduce an `Enum` or something similar for these kind of values in the main code and use that in the test, too. I know the code was like this before, but unfortunately it is not really organised well. This is the case when we can use the `boy-scout` rule. Leave everything in a better way than you found it.

ke4 (changes requested):
Added 2 change requests. Thanks

app/src/test/java/uk/ac/ebi/atlas/experiments/ExperimentDesignControllerWIT.java
lingyun1010: If you mean the list of `"contrastName", "referenceOrTest", "run"`, they are only tested in this class, so I think it is fine to leave it like that. If we would like to improve/refactor the code somehow, we may raise another story regarding that.

app/src/test/java/uk/ac/ebi/atlas/experiments/ExperimentDesignControllerWIT.java
ke4: This test class has been removed already

app/src/test/java/uk/ac/ebi/atlas/experiments/ExperimentDesignControllerWIT.java
ke4: This test class has been removed already

app/src/main/java/uk/ac/ebi/atlas/experimentpage/tabs/ExperimentPageContentSerializer.java
ke4: This code fragment has been removed as it was no longer needed.

app/src/main/java/uk/ac/ebi/atlas/experiments/ExperimentDesignController.java
ke4: This class has been removed as it was not needed anymore.

app/src/main/java/uk/ac/ebi/atlas/experimentpage/tabs/ExperimentPageContentSerializer.java
ke4: This code fragment has been removed already as it was not needed anymore.

app/src/test/java/uk/ac/ebi/atlas/experiments/ExperimentDesignControllerWIT.java
ke4: This test class has been removed already as its test subject was not needed anymore.

ke4 (approved):
Look good to me

PR #332, opened 2023-03-15
Feature/validate parameterisation for cell type endpoints

lingyun1010:
In this feature, we check if the user-input parameterisation value is existed in the database.

For example: `http://localhost:8080/gxa/sc/json/cell-plots/E-EHCA-2/clusters/metadata/inferred%20cell%20type%20-%20authors%20labels?plotMethod=tsne&accessKey=&perplexity=12` should return 
```
error | "Invalid plot parameter value perplexity=12"
```


upendrakumbham (approved):
ALGTM!

PR #333, opened 2023-03-15
Chore – Gene search optimization

alfonsomunozpomer:
In a nutshell:
- Gene search results have been trimmed and only experiments where the searched genes are marker genes are returned
- *Homo sapiens* is the default species to avoid timeouts or excessive waiting for the majority of users
- The Ensembl ID example has been changed to one that won’t error out (the previous one would work after caching)

Besides some tweaks and fixes that I didn’t want to get lost (more specifically changes in our Docker files).

### DIFF
--- a/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
@@ -83,10 +83,10 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
         var resultsBuilder = ImmutableList.builder();
         // This was a very expressive stream chain (look it up in the history), but to filter by results where the gene
         // is a marker, making it a for loop is more readable IMHO; I also noticed a 10% performance improvement
-        for (var i = 0; i < expressedGeneIdEntries.size(); i++) {
-            var geneId = expressedGeneIdEntries.get(i).getKey();
+        for (var expressedGeneIdEntry : expressedGeneIdEntries) {
+            var geneId = expressedGeneIdEntry.getKey();
 
-            var exp2cellsIt = expressedGeneIdEntries.get(i).getValue().entrySet().iterator();
+            var exp2cellsIt = expressedGeneIdEntry.getValue().entrySet().iterator();
             while (exp2cellsIt.hasNext()) {
                 var exp2cells = exp2cellsIt.next();
                 var experimentAccession = exp2cells.getKey();
--- a/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
@@ -86,9 +86,7 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
         for (var expressedGeneIdEntry : expressedGeneIdEntries) {
             var geneId = expressedGeneIdEntry.getKey();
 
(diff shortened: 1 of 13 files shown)


app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -46,78 +51,104 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
 
     @GetMapping(value = "/json/search", produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
     public String search(@RequestParam MultiValueMap<String, String> requestParams) {
+        var stopWatch = new StopWatch();
```
ke4, line 54: Would you like to deploy all of this measurements to public?
If yes, than it should be temporary, I guess.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -46,78 +51,104 @@ public class JsonGeneSearchController extends JsonExceptionHandlingController {
...
+        var resultsBuilder = ImmutableList.builder();
+        // This was a very expressive stream chain (look it up in the history), but to filter by results where the gene
+        // is a marker, making it a for loop is more readable IMHO; I also noticed a 10% performance improvement
+        for (var i = 0; i < expressedGeneIdEntries.size(); i++) {
+            var geneId = expressedGeneIdEntries.get(i).getKey();
+
+            var exp2cellsIt = expressedGeneIdEntries.get(i).getValue().entrySet().iterator();
+            while (exp2cellsIt.hasNext()) {
+                var exp2cells = exp2cellsIt.next();
+                var experimentAccession = exp2cells.getKey();
```
ke4: I suggest to use the for-each loop instead of the classic for/while loop, as you don't really need the `i` variable and it is more readable and more importantly!!! ;-) less code to maintain

```suggestion
       for (var geneIdEntry : expressedGeneIdEntries) {
            var geneId = geneIdEntry.getKey();

            for (var exp2cells: geneIdEntry.getValue().entrySet()) {
                var experimentAccession = exp2cells.getKey();
```

ke4 (changes requested):
Looks good, but I requested a small change. Could you check it, please.
If I am very strict I would say the docker changes belongs to another PR, but I closed my eyes for now. 😉
Thanks for your hard work 👍

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
alfonsomunozpomer: I’m unsure of what to do here, tbh. Stopwatch impact is negligible and logging the results as a debug message will mean that it won’t have any effect in public/fallback environments. This is not there in the `develop` branch. I guess having two code paths for performance and another one for profiling is too much?

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
alfonsomunozpomer: Totally true. I’m so used to using streams that I had forgotten about the auto-increment notation. :crying_cat_face:

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: It is fine. We can leave it here.

ke4 (approved):
LGTM 👍

Issue #278, opened 2022-08-04
Too much recursion error when searching for Gallus gallus metadata


alfonsomunozpomer:
https://www.ebi.ac.uk/gxa/sc/search/metadata/Gallus%20gallus

The error happens whenever you search for a species in the metadata search box.
For example: liver.
Species are not passes to the controller.

upendrakumbham:
Created PR for this issue, hence closing the issue.

PR #334, opened 2023-03-16
Fix json payload error which causes recursion error in the front end …

upendrakumbham:
Issue reference: [Recursion error](https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/278)

Getting recursion error in the front end because of the JSON response error for the metadata 'Gallus gallus'. In the JSON response tree, one element generates the same value for the "ID" and "Parent" attributes.

Recursion error URL:  https://www.ebi.ac.uk/gxa/sc/search/metadata/Gallus%20gallus

Endpoint - ` https://www.ebi.ac.uk/gxa/sc/json/cell-type-wheel/Gallus gallus`

Fix: Tweaked a code to generate JSON payload attribute values "ID" and "Parent" are always different.


alfonsomunozpomer (changes requested):
Please add a test that fails with the faulty code and passes after adding your fix.

app/src/test/java/uk/ac/ebi/atlas/search/metadata/HighchartsSunburstAdapterTest.java
```diff
@@ -0,0 +1,110 @@
...
+                                        species1.getName(), species1OrganismPart1, species1OrganismPart1CellType3),
+                                species1OrganismPart1ExperimentAccession)),
+                        // Organism part 2, cell types
+                        ImmutableSet.of(ImmutablePair.of(
+                                ImmutableList.of(
+                                        species1.getName(), species1OrganismPart2, species1OrganismPart2CellType1),
+                                species1OrganismPart2ExperimentAccession)));
+    }
+
+    @DisplayName("This unit test use the randomised input test data and test the functionality of the wheel")
```
alfonsomunozpomer: Rephrase without words such as “This unit test...” or “This test...” (see below). This is more of a code comment rather than a test description, since it mentions implementation details. The `@DisplayName` should be a human-readable firendlier version of the test method; for instance, something like:
> Cell type wheel should not have the same value for the ID and its parent

app/src/test/java/uk/ac/ebi/atlas/search/metadata/HighchartsSunburstAdapterTest.java
```diff
@@ -0,0 +1,110 @@
...
+    @Test
+    void cellTypeWheelSunburstResultObjectShouldNotContainIdAndParentAttributeValuesSame() {
+        ImmutableSet<ImmutableMap<String, ?>> result = subject.getCellTypeWheelSunburst(metadataSearchTerm,cellTypeWheelServiceMock
```
alfonsomunozpomer: You can use `var` here.

app/src/test/java/uk/ac/ebi/atlas/search/metadata/HighchartsSunburstAdapterTest.java
```diff
@@ -0,0 +1,110 @@
...
+                                ImmutablePair.of(
+                                        ImmutableList.of("Gallus gallus"), ("E-CURD-13"))),
+                        ImmutableSet.of(
+                                ImmutablePair.of(
+                                        ImmutableList.of("Gallus gallus"), ("E-CURD-12"))),
+                        ImmutableSet.of(
+                                ImmutablePair.of(
+                                        ImmutableList.of("Gallus gallus"), ("E-GEOD-89910"))));
+
+        ImmutableSet<ImmutableMap<String, ?>> result = subject.getCellTypeWheelSunburst("Gallus gallus",cellTypeWheelServiceMock
```
alfonsomunozpomer: Use `var` again.

app/src/test/java/uk/ac/ebi/atlas/search/metadata/HighchartsSunburstAdapterTest.java
```diff
@@ -0,0 +1,110 @@
...
+        ImmutableSet<ImmutableMap<String, ?>> result = subject.getCellTypeWheelSunburst(metadataSearchTerm,cellTypeWheelServiceMock
+                .search(metadataSearchTerm, species1.getName()));
+        assertThat(result).hasSize(2);
+        assertThat(result.asList().get(0).get("id")).isNotEqualTo(result.asList().get(0).get("parent"));
+        assertThat(result.asList().get(1).get("id")).isNotEqualTo(result.asList().get(1).get("parent"));
+    }
+
+    @Test
+    @DisplayName("This test uses hardcoded mock test data produced by the public server and tests the pre and post-bugfix code precisely to make sure that bug will not reappear again." +
+            "Both unit tests look the same, the only difference is in the input test data.")
```
alfonsomunozpomer: Again, these are code comments. A unit test should’t mention other tests for understanding its meaning.

alfonsomunozpomer (changes requested):
Some changes in wording and style. Other than that it’s mostly ok.

alfonsomunozpomer (changes requested):
Some changes in wording and style. Other than that it’s mostly ok.

app/src/test/java/uk/ac/ebi/atlas/search/metadata/HighchartsSunburstAdapterTest.java
```diff
@@ -0,0 +1,114 @@
...
+    void cellTypeWheelSunburstResultObjectShouldNotContainIdAndParentAttributeValuesSame() {
+        var result = subject.getCellTypeWheelSunburst(metadataSearchTerm,cellTypeWheelServiceMock
+                .search(metadataSearchTerm, species1.getName()));
+        assertThat(result).hasSize(2);
+        assertThat(result.asList().get(0).get("id")).isNotEqualTo(result.asList().get(0).get("parent"));
+        assertThat(result.asList().get(1).get("id")).isNotEqualTo(result.asList().get(1).get("parent"));
+    }
+
+    @Test
+    void ShouldThrowAssertionErrorIfWheelResultIdAndParentAttributeValuesAreSame() {
```
alfonsomunozpomer: Whenever a test fails, it throws an assertion error/exception, so this shouldn’t be part of what you’re testing. In other words, a test that is named:
```java
@Test
void throwsAssertionErrorIfTheSubjectWorksUnderSomeCondition() {
// ...
}
```

Should be renamed to:
```java
void subjectWorksUnderSomeCondition() {
// ...
}
```

One last observation: what’s the difference between this test and the previous one? What are you testing in each case? That should be your test name.

alfonsomunozpomer (changes requested):
Some clarification for the tests required.

app/src/test/java/uk/ac/ebi/atlas/search/metadata/HighchartsSunburstAdapterTest.java
```diff
@@ -75,18 +76,17 @@ void setUp() {
...
-        assertThat(result.asList().get(0).get("id")).isNotEqualTo(result.asList().get(0).get("parent"));
-        assertThat(result.asList().get(1).get("id")).isNotEqualTo(result.asList().get(1).get("parent"));
+        assertFalse(result.isEmpty());
+        assertNotEquals("", result.asList().get(1).get("id"));
+        assertNotEquals("", result.asList().get(1).get("parent"));
     }
 
     @Test
-    void ShouldThrowAssertionErrorIfWheelResultIdAndParentAttributeValuesAreSame() {
+    void failsIfIdAndParentFieldValuesAreSame() {
```
alfonsomunozpomer: Again, “fails” isn’t what you’re testing. Your test name describes the conditions and/or expectations of good behaviour. If it helps, @ke4 sent this a while ago to our Slack channel: https://osherove.com/blog/2005/4/3/naming-standards-for-unit-tests.html.

alfonsomunozpomer (approved):
All good!

PR #337, opened 2023-03-21
Bugfix – Solr health endpoint reports Solr to be down even though replicas are active

alfonsomunozpomer:
Our `SolrCloudHealthService` was fine when we had no replicas, but that’s no longer the case. Besides the change in the logic I needed to update Mockito to v5.2 in order to be able to mock `final` methods.

I also took the chance to fix some other minor issues I came across.

I leave as a reference an example value of a response to [the `CLUSTER_STATUS` request](https://solr.apache.org/guide/8_7/cluster-node-management.html#clusterstatus) viewed through the lens of SolrJ.

clusterstatus


app/src/test/java/uk/ac/ebi/atlas/search/GeneSearchServiceTest.java
ke4, line 1: Well, although it is a code improvement, but this does not belong to Solr health endpoint improvement.
Next time, please add this kind of not related changes to another PR.

docker/dev-test.env
```diff
@@ -9,7 +9,7 @@ POSTGRES_HOST=scxa-postgres-test
 POSTGRES_DB=gxpscxadev
 POSTGRES_USER=atlasdev
 POSTGRES_PASSWORD=atlasdev
-SCHEMA_VERSION=18
+SCHEMA_VERSION=latest
```
ke4, line 12: Are we using `anndata` now?

ke4 (approved):
I am going to approve this PR, but next time it would be better to add unrelated code changes to another PR. Thanks

app/src/test/java/uk/ac/ebi/atlas/search/GeneSearchServiceTest.java
alfonsomunozpomer: Change in L201 was required because of the change in Mockito version. The rest was some cleaning up of the same class. I don’t think it’s worth to have PRs that only have changes of variable declarations and it’s better to do that as part of the tasks we do on a regular basis. We can discuss how we want to do it in any case. Sorry for the added noise.

docker/dev-test.env
alfonsomunozpomer: We are in `develop`!

PR #341, opened 2023-05-02
Fix default plot type and it's parameterisation null check code issues

upendrakumbham:
Fixed a few code issues:

1.  Added a check in the `getMiddleElement(List plotOptions)` method
2. Removed hardcoded  cell plot methods in the method `fetchDefaultPlotMethodWithParameterisation(String experimentAccession) `


app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
```diff
@@ -103,27 +103,27 @@ public ImmutableSet<TSnePoint> expressionPlot(String experimentAccession,
...
-
-        if (plotOptions.size() % 2 == 0) { // even number
-            Object umapEvenItem = plotOptions.get((plotOptions.size() / 2 - 1));
-            return (JsonObject) umapEvenItem;
-        } else { //odd number
-            Object umapOddItem = plotOptions.get((plotOptions.size() / 2));
-            return (JsonObject) umapOddItem;
-        }
+    private JsonObject getMiddleElement(List plotOptions) {
+        if (!plotOptions.isEmpty() && plotOptions.size() > 0) {
```
ke4: We might also need a `null` check here. What do you think?

ke4 (changes requested):
I thought we would need a test case for the original bug, but I can't find any.
What do you think?
I also added a comment.
Thanks

app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
upendrakumbham: Yes, as a safe side.

alfonsomunozpomer:
Please add tests!

upendrakumbham:
> Please add tests!

We have tests for this @alfonsomunozpomer, positive and negative tests. Are you expecting any specific case like NullPointerException coverage for the code you mean?

upendrakumbham:
Let's chat on this after standup to add more test coverage.

app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
```diff
@@ -103,27 +103,27 @@ public ImmutableSet<TSnePoint> expressionPlot(String experimentAccession,
...
+            if (plotOptions.size() % 2 == 0) { // even number
+                Object umapEvenItem = plotOptions.get((plotOptions.size() / 2 - 1));
+                return (JsonObject) umapEvenItem;
+            } else { //odd number
+                Object umapOddItem = plotOptions.get((plotOptions.size() / 2));
+                return (JsonObject) umapOddItem;
+            }
+        } else
+            return new JsonObject();
```
alfonsomunozpomer: **ALWAYS** enclose `if` and `else` clauses in curly braces!

alfonsomunozpomer (changes requested):
Besides the `else` clause, remember to design a test that fails without the fix, and passes with the fix.

app/src/test/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotServiceTest.java
```diff
@@ -168,8 +171,6 @@ void throwsNullPointerExceptionIfHardCodedTestMethodIsNotSameAsDBMethod() {
                         List.of(new Gson().fromJson("{\"perplexity\": 20}", JsonObject.class))));
 
         assertThrows(NullPointerException.class, () -> {
-            //We are passing method 'umap' which is small letters, But DB has capilized method 'UMAP'
-            //This gives null List as MAP fetches from the DB, MAP keys are case-sensitive.
             subject.fetchDefaultPlotMethodWithParameterisation(experimentAccession)
                     .get(umap).getAsJsonObject();
         });
```
alfonsomunozpomer: This is a bad test. The method `CellPlotService::fetchDefaultPlotMethodWithParameterisation` returns an `ImmutableMap`, which will throw an NPE if you try to get a key which does not exist. This isn’t testing any business logic, but Guava’s `ImmutableMap` implementation.

alfonsomunozpomer (changes requested):
Please see my comment on the modified test. It’s worth noting that the test in the original PR was also inadequate. The important point here is to realise we’re testing something which is 3rd party code (the `ImmutableMap::get` method with a random value). If there’s no way to reformulate the test (i.e. are there any values you expect the map to have?) then remove it.

ke4 (commented):
In the `JdbcUtils` there is a new test method that is not being used. If that is true, then shall we delete it, please?

ke4 (approved):

alfonsomunozpomer (approved):
All good. Merge away! :)

PR #342, opened 2023-05-04
bugfix/default plot method and parameterisation

upendrakumbham:
Fixed a few code issues:

- Added a check in the getMiddleElement(List plot option) method

- Removed hardcoded cell plot methods in the method fetchDefaultPlotMethodWithParameterisation(String experimentAccession)

### DIFF
--- a/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
@@ -115,7 +115,7 @@ public class CellPlotService {
 
     private JsonObject getMiddleElement(List plotOptions) {
 
-        if (plotOptions.isEmpty() || plotOptions == null) {
+        if (plotOptions == null || plotOptions.isEmpty()) {
             return new JsonObject();
         } else {
             if (plotOptions.size() % 2 == 0) { // even number
--- a/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
@@ -118,13 +118,9 @@ public class CellPlotService {
         if (plotOptions == null || plotOptions.isEmpty()) {
             return new JsonObject();
         } else {
-            if (plotOptions.size() % 2 == 0) { // even number
-                Object umapEvenItem = plotOptions.get((plotOptions.size() / 2 - 1));
-                return (JsonObject) umapEvenItem;
-            } else { //odd number
-                Object umapOddItem = plotOptions.get((plotOptions.size() / 2));
-                return (JsonObject) umapOddItem;
-            }
+            var plotOptionsSize = plotOptions.size();
+            var middleOfPlotOptionsSize = plotOptionsSize % 2 == 0 ? plotOptionsSize / 2 - 1 : plotOptionsSize / 2;
+            return (JsonObject) plotOptions.get(middleOfPlotOptionsSize);
         }
 
     }
--- a/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
@@ -114,7 +114,10 @@ public class CellPlotService {
     }
 
     private JsonObject getMiddleElement(List plotOptions) {
-        if (!plotOptions.isEmpty() && plotOptions.size() > 0) {
+
+        if (plotOptions.isEmpty() || plotOptions == null) {
+            return new JsonObject();
+        } else {
             if (plotOptions.size() % 2 == 0) { // even number
                 Object umapEvenItem = plotOptions.get((plotOptions.size() / 2 - 1));
                 return (JsonObject) umapEvenItem;
@@ -122,8 +125,8 @@ public class CellPlotService {
                 Object umapOddItem = plotOptions.get((plotOptions.size() / 2));
                 return (JsonObject) umapOddItem;
             }
-        } else
-            return new JsonObject();
+        }
+
     }
 
 }
--- a/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
@@ -114,15 +114,9 @@ public class CellPlotService {
     }
 
     private JsonObject getMiddleElement(List plotOptions) {
-
-        if (plotOptions == null || plotOptions.isEmpty()) {
-            return new JsonObject();
-        } else {
             var plotOptionsSize = plotOptions.size();
             var middleOfPlotOptionsSize = plotOptionsSize % 2 == 0 ? plotOptionsSize / 2 - 1 : plotOptionsSize / 2;
             return (JsonObject) plotOptions.get(middleOfPlotOptionsSize);
-        }
-
     }
 
 }
--- a/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
@@ -104,11 +104,11 @@ public class CellPlotService {
 
     public ImmutableMap<String, JsonObject> fetchDefaultPlotMethodWithParameterisation(String experimentAccession) {
 
-        var defaultcellPlots = cellPlotDao.fetchDefaultPlotMethodWithParameterisation(experimentAccession);
+        var defaultCellPlots = cellPlotDao.fetchDefaultPlotMethodWithParameterisation(experimentAccession);
 
         ImmutableMap.Builder<String, JsonObject> defaultPlotTypeAndOptions = new ImmutableMap.Builder<>();
 
-        defaultcellPlots.forEach((method, options) -> defaultPlotTypeAndOptions.put(method, getMiddleElement(defaultcellPlots.get(method))));
+        defaultCellPlots.forEach((method, options) -> defaultPlotTypeAndOptions.put(method, getMiddleElement(defaultCellPlots.get(method))));
 
         return defaultPlotTypeAndOptions.build();
     }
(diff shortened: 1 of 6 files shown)


app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
```diff
@@ -103,27 +103,30 @@ public ImmutableSet<TSnePoint> expressionPlot(String experimentAccession,
...
-
-        if (plotOptions.size() % 2 == 0) { // even number
-            Object umapEvenItem = plotOptions.get((plotOptions.size() / 2 - 1));
-            return (JsonObject) umapEvenItem;
-        } else { //odd number
-            Object umapOddItem = plotOptions.get((plotOptions.size() / 2));
-            return (JsonObject) umapOddItem;
+    private JsonObject getMiddleElement(List plotOptions) {
+
+        if (plotOptions.isEmpty() || plotOptions == null) {
```
ke4: if `plotOptions` is `null` this code is going to fail as you can't call a method on a null object, in this case `isEmpty()`. Please change the condition as I suggest above. You can just apply my suggestion.

```suggestion
        if (plotOptions == null || plotOptions.isEmpty()) {
```

app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
```diff
@@ -103,27 +103,30 @@ public ImmutableSet<TSnePoint> expressionPlot(String experimentAccession,
...
+            return new JsonObject();
+        } else {
+            if (plotOptions.size() % 2 == 0) { // even number
+                Object umapEvenItem = plotOptions.get((plotOptions.size() / 2 - 1));
+                return (JsonObject) umapEvenItem;
+            } else { //odd number
+                Object umapOddItem = plotOptions.get((plotOptions.size() / 2));
+                return (JsonObject) umapOddItem;
+            }
```
ke4:
```suggestion
            var plotOptionsSize = plotOptions.size();
            var middleOfPlotOptionsSize = plotOptionsSize % 2 == 0 ? plotOptionsSize / 2 - 1 : plotOptionsSize / 2;
            return (JsonObject) plotOptions.get(middleOfPlotOptionsSize);
```

app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
```diff
@@ -103,27 +103,30 @@ public ImmutableSet<TSnePoint> expressionPlot(String experimentAccession,
...
-        if (plotOptions.size() % 2 == 0) { // even number
-            Object umapEvenItem = plotOptions.get((plotOptions.size() / 2 - 1));
-            return (JsonObject) umapEvenItem;
-        } else { //odd number
-            Object umapOddItem = plotOptions.get((plotOptions.size() / 2));
-            return (JsonObject) umapOddItem;
+    private JsonObject getMiddleElement(List plotOptions) {
+
+        if (plotOptions.isEmpty() || plotOptions == null) {
+            return new JsonObject();
```
ke4: Is an empty `JsonObject` is a valid return value here?
While I am not sure about the answer I won't check the tests.
That's one of the reason that it is better to write a test/tests first, but it looks like you wrote the tests later. Am I correct?

ke4 (changes requested):
I added some comments. Could you review them please?
Thanks

app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
upendrakumbham: I fixed your comments @ke4. Yes, correct,  I fixed the tests after implementation. Now, I managed to assert an Empty JSON string.

I'm also not happy with that empty JSON object :).

How about throwing a custom exception when we met this condition `if (plotOptions == null || plotOptions.isEmpty()) {`

ke4: I think it is probably better to throw an exception, but we also need to think if that could happen at all OR if with a refactor we could eliminate this case.

app/src/test/java/uk/ac/ebi/atlas/experimentpage/TSnePlotSettingsServiceIT.java
```diff
@@ -163,6 +161,8 @@ private long getOpenFileCount() {
     }
 
     private Stream<String> randomSingleCellExperimentAccessionProvider() {
```
ke4: The name of the method is not reflecting the result that it is returning.  Could you change the name of the method, please?
Could you also fix the formatting, please?

app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotService.java
```diff
@@ -103,27 +103,20 @@ public ImmutableSet<TSnePoint> expressionPlot(String experimentAccession,
...
+        var defaultCellPlots = cellPlotDao.fetchDefaultPlotMethodWithParameterisation(experimentAccession);
 
         ImmutableMap.Builder<String, JsonObject> defaultPlotTypeAndOptions = new ImmutableMap.Builder<>();
-        if (!defaultcellPlots.isEmpty()) {
-            defaultPlotTypeAndOptions.put(umap, getMiddleElement(defaultcellPlots.get(umap)));
-            defaultPlotTypeAndOptions.put(tsne, getMiddleElement(defaultcellPlots.get(tsne)));
-        }
+
+        defaultCellPlots.forEach((method, options) -> defaultPlotTypeAndOptions.put(method, getMiddleElement(defaultCellPlots.get(method))));
+
```
ke4: How about this code:

```
ImmutableMap.Builder<String, JsonObject> defaultPlotTypeAndOptions = new ImmutableMap.Builder<>();
cellPlotDao.fetchDefaultPlotMethodWithParameterisation(experimentAccession)
    .forEach((method, options) -> defaultPlotTypeAndOptions.put(method, getMiddleElement(options)));
````

File: app/src/test/java/uk/ac/ebi/atlas/experimentpage/TSnePlotSettingsServiceIT.java [hunk 6]
```diff
@@ -141,10 +141,9 @@ void filesClosed(String experimentAccession) {
 
     @ParameterizedTest
     @MethodSource("randomSingleCellExperimentAccessionProvider")
-    void getTSnePlotTypesAndOptionsForValidAccession(String experimentAccession) {
+    void getTSnePlotTypesAndOptionsForValidAccession(String experimentAccession,String plotMethod) {
```
ke4:
```suggestion
    void getTSnePlotTypesAndOptionsForValidAccession(String experimentAccession, String plotMethod) {
```

app/src/test/java/uk/ac/ebi/atlas/experimentpage/TSnePlotSettingsServiceIT.java
```diff
@@ -162,7 +161,9 @@ private long getOpenFileCount() {
         }
     }
 
-    private Stream<String> randomSingleCellExperimentAccessionProvider() {
-        return Stream.of(jdbcTestUtils.fetchRandomExperimentAccession());
+    private Stream<Arguments> randomSingleCellExperimentAccessionProvider() {
```
ke4: The method name is not reflecting what the method is doing. It is not only returning a random experiment accession, but also a plot method.
Please, either extract that other functionality to another helper method or change the name of the method.
Thanks

app/src/test/java/uk/ac/ebi/atlas/experimentpage/TSnePlotSettingsServiceIT.java
```diff
@@ -162,7 +161,9 @@ private long getOpenFileCount() {
...
+        var experimentAccession = jdbcTestUtils.fetchRandomExperimentAccession();
+        var plotMethod = jdbcTestUtils.fetchRandomPlotMethod(experimentAccession);
+        return Stream.of(Arguments.of(experimentAccession,plotMethod));
```
ke4, line 167:
```suggestion
        return Stream.of(Arguments.of(experimentAccession, plotMethod));
```

app/src/test/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotServiceTest.java
```diff
@@ -129,24 +130,40 @@ void getCellPlotParameter() {
     }
 
     @Test
-    void fetchDefaultPlotMethodWithParameterisation(){
+    void fetchDefaultPlotMethodWithParameterisation() {
         when(cellPlotDaoMock.fetchDefaultPlotMethodWithParameterisation("E-CURD-4"))
-                .thenReturn(ImmutableMap.of("umap",
+                .thenReturn(ImmutableMap.of("UMAP",
```
ke4: I think the last time we discussed that we are going to generate a random string as the name of the plot type and accession.
I also think that you have the helper method for that. Could you please use that method here, too?
Thanks

app/src/test/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotServiceTest.java
```diff
@@ -129,24 +130,40 @@ void getCellPlotParameter() {
...
+    void returnEmptyResultIfThereIsNoDefaultPlotMethodAndParameterisation() {
         when(cellPlotDaoMock.fetchDefaultPlotMethodWithParameterisation("fooBar"))
                 .thenReturn(ImmutableMap.of());
 
         assertThat(subject.fetchDefaultPlotMethodWithParameterisation("fooBar"))
                 .isEmpty();
     }
+
+    @Test
+    void throwsNullPointerExceptionIfHardCodedTestMethodIsNotSameAsDBMethod() {
```
ke4, line 160: The same comment as in line 135 (randomised strings). Thanks

ke4 (changes requested):
I added comments for requesting some changes. Could you please review them.
Many thanks, Karoly

app/src/test/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotServiceTest.java
upendrakumbham: I added randomised strings and accessions for the plot type and experiment accession.

upendrakumbham: thanks for the comment.

ke4 (approved):
LGTM
Thanks for your work @upendrakumbham

PR #346, opened 2023-05-17
Enable atlas-web-core integration tests in Single Cell Expression Atlas Gradle container

alfonsomunozpomer:
> We’ll need `magetab` and experiment designs for bulk, possibly Solr, too. This could make the `execute-all-tests.sh` script obsolete.

Summary of changes to the `prepare-dev-environment` directory:
- All scripts have been changed to use `docker compose` instead of Docker. I described the new requirements in `README.md` and removed the “Troubleshooting” section since that only happened with earlier versions of Docker and the deprecated `docker-compose`.
- I added a `-r` flag to all scripts in order to remove volumes. Being able to explicitly mark which volumes shouldn’t be deleted with `external:true` when executing `docker compose down -v` is quite handy.
- I added an interactive countdown before removing volumes. I asked Copilot for help with that one and I tweaked it afterwards. Good job Copilot!
- The Solr private key is used by the populator container as a secret, rather than a volume.

For the Compose files that we use to launch and test the application:
- Got rid of the error you had some times regarding the signing key in Solr. The service only needs the key when verifying packages (i.e. when running the `prepare-dev-environment` script). If it’s not set it defaults to `/dev/null` and it’s fine.
- I added a new `docker-compose-build.yml` to be used in conjunction with the Tomcat service. It will build the front-end packages, the WAR, **and then** launch Tomcat. Use the `build-and-deploy-webapp.sh` script to do that. You get a `.war` file as a side benefit.

This is all as a result of the need to parameterise all the YAML files with `PROJECT_NAME` to dynamically change between `gxa` and `scxa`. It was a massive tangent. The positive side of this is that we may be able to use exactly the same files for bulk and SC, but this needs some consideration.

There are some additional changes to test classes and a few other main classes in `atlas-web-core`, too. See https://github.com/ebi-gene-expression-group/atlas-web-core/pull/128.

Remember that this PR makes it possible to run integration tests for `atlas-web-core` locally with Docker. Jenkins doesn’t do that, because we don’t have a bulk CI environemnt yet. One exmple output of `execute-all-tests.sh` after adding integration tests in `atlas-web-core` is here:
https://gist.github.com/alfonsomunozpomer/61ec9437124d93e672307813ad596b61


app/src/test/java/uk/ac/ebi/atlas/experiments/ExperimentCellCountDaoImplIT.java
```diff
@@ -0,0 +1,74 @@
...
+    void setUp() {
+        this.subject = new ExperimentCellCountDaoImpl(namedParameterJdbcTemplate);
+    }
+
+    @Test
+    void testForNumberOfCells() {
+        // Most experiments have 50 cells, but there’s at least one with 51
+        // These numbers correspond to the AnnData fixtures:
+        // https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/306
+        assertThat(subject.fetchNumberOfCellsByExperimentAccession(jdbcTestUtils.fetchRandomExperimentAccession()))
```
ke4: If we add a new experiment with different numbers of cells, then this test is going to be broken.
I am not sure whether this test adds any value to our test suite.

build-and-deploy-webapp.sh
ke4, line 1: 👍 Cheers for this one! Going to be very useful! 👍

docker/docker-compose-build.yml
```diff
@@ -0,0 +1,43 @@
+version: "3.6"
+
+services:
+  node-webpack:
+    image: node:16
+    volumes:
+      - ..:/root/project
+    working_dir: /root/project
+    command: ["sh", "-c", "npm install -g npm-check-updates && ./compile-front-end-packages.sh -ui" ]
```
ke4, line 9: We could rewrite this using the `|` operator as in `bulk`. The code would be a bit cleaner.

docker/docker-compose-gradle.yml
```diff
@@ -1,65 +1,45 @@
...
-      - scxa-solrcloud-1
-      - scxa-flyway-test
+      solrcloud-0:
+        condition: service_started
+      solrcloud-1:
+        condition: service_started
+      flyway:
+        condition: service_completed_successfully
     environment:
       POSTGRES_HOST: $POSTGRES_HOST
```
ke4: We don't need to redefine the environment variables that was already defined and sourced in an environment file.

docker/docker-compose-postgres-test.yml
```diff
@@ -1,38 +1,42 @@
...
     image: flyway/flyway
-    container_name: scxa-flyway-test
     networks:
       - atlas-test-net
+    depends_on:
+      - postgres-test
+    volumes:
+      - ../schemas/flyway/${PROJECT_NAME}:/flyway/sql
+    environment:
+      POSTGRES_HOST: ${POSTGRES_HOST}
```
ke4: We don't need to redefine the environment variables that was already defined and sourced in an environment file.

docker/docker-compose-postgres.yml
```diff
@@ -1,44 +1,53 @@
...
 services:
-  scxa-postgres:
-    image: postgres:11-alpine
+  postgres:
     container_name: ${POSTGRES_HOST}
+    image: postgres:11-alpine
+    networks:
+      - atlas-test-net
     environment:
       POSTGRES_DB: ${POSTGRES_DB}
```
ke4: We don't need to redefine the environment variables that was already defined and sourced in an environment file.

docker/prepare-dev-environment/gradle-cache/run.sh
```diff
@@ -2,28 +2,38 @@
 set -e
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 
-# GRADLE_DISTS_VOL_NAME
-# GRADLE_RO_DEP_CACHE_VOL_NAME
-source ${SCRIPT_DIR}/../../dev.env
+#PROJECT_NAME=scxa
```
ke4, line 5: Could we remove these commented env variable definitions, please?

docker/prepare-dev-environment/postgres/docker-compose.yml
```diff
@@ -0,0 +1,105 @@
...
+      flyway:
+        condition: service_completed_successfully
+    volumes:
+      - gradle-wrapper-dists:/root/.gradle/wrapper/dists
+      - gradle-ro-dep-cache:/gradle-ro-dep-cache:ro
+      - atlas-data-exp:/atlas-data/exp
+      - atlas-data-expdesign:/atlas-data/expdesign
+    environment:
+      JAVA_TOOL_OPTIONS: "-Dfile.encoding=UTF8"
+      EXP_IDS: ${EXP_IDS}
```
ke4: We don't need to redefine the environment variables that was already defined and sourced in an environment file.

docker/prepare-dev-environment/solr/Dockerfile
```diff
@@ -1,22 +1,22 @@
...
+RUN apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+
+# Create the necessary directory
 RUN mkdir -p /root/bioentity-properties-jsonl
 
+# Clone the necessary repositories
 WORKDIR /root
 RUN git clone --depth 1 https://github.com/ebi-gene-expression-group/index-bioentities.git
 RUN git clone --depth 1 --recurse-submodules https://github.com/ebi-gene-expression-group/atlas-web-bulk.git
```
ke4, line 19: Is it intentionally `atlas-web-bulk` or it should be `atlas-web-single-cell`?

docker/prepare-dev-environment/solr/docker-compose.yml
```diff
@@ -0,0 +1,128 @@
+version: "3.6"
+
+services:
+  postgres-stub:
+    container_name: ${POSTGRES_HOST}
+    image: postgres:11-alpine
+    networks:
+      - atlas-test-net
+    environment:
+      POSTGRES_DB: ${POSTGRES_DB}
```
ke4: We don't need to redefine the environment variables that was already defined and sourced in an environment file.

docker/prepare-dev-environment/solr/docker-compose.yml
```diff
@@ -0,0 +1,128 @@
...
+      - gradle-wrapper-dists:/root/.gradle/wrapper/dists
+      - gradle-ro-dep-cache:/gradle-ro-dep-cache:ro
+      - atlas-data-bioentity-properties:/atlas-data/bioentity_properties
+      - atlas-data-exp:/atlas-data/exp
+      - atlas-data-expdesign:/atlas-data/expdesign
+    secrets:
+      - solrcloud.pem
+    environment:
+      JAVA_TOOL_OPTIONS: "-Dfile.encoding=UTF8"
+      EXP_IDS: ${EXP_IDS}
```
ke4: We don't need to redefine the environment variables that was already defined and sourced in an environment file.

docker/prepare-dev-environment/solr/docker-compose.yml
```diff
@@ -0,0 +1,128 @@
...
+      SOLR_COLLECTION_BIOENTITIES_SCHEMA_VERSION: 1
+    working_dir: /root
+    command:
+      - |
+        for _SPECIES in $${SPECIES}
+        do
+          rsync -av --include=$${_SPECIES}* --include=*/ --exclude=* /atlas-data/bioentity_properties /root
+        done
+  
+        cd /root/atlas-web-bulk
```
ke4, line 48: Are we getting the `bioentities` JSON files from `bulk` for `scxa`?

ke4 (dismissed):
Great work! 👍 
I added similar comments as in my previous review for your other PR.
Many thanks

docker/prepare-dev-environment/solr/Dockerfile
alfonsomunozpomer: It’s intentional: the bulk CLI is the one that has the commands to transform gene annotations into JSONL files suitable to be ingested by Solr.

alfonsomunozpomer:
@ke4 Please have another look. You were right and variables that are in --env-file aren’t necessary to be declared another time in the Docker Compose YAML files.

docker/docker-compose-gradle.yml
alfonsomunozpomer: Removed!

docker/docker-compose-postgres-test.yml
alfonsomunozpomer: Removed!

docker/prepare-dev-environment/solr/docker-compose.yml
alfonsomunozpomer: Removed!

docker/prepare-dev-environment/solr/docker-compose.yml
alfonsomunozpomer: Removed!

docker/prepare-dev-environment/solr/docker-compose.yml
alfonsomunozpomer: From neither, these come from external services and we keep them as TSVs in a project-agnostic location. We use then the CLI in `atlas-web-bulk` to transform them into JSONL before sending them to the service-specific Solr.

app/src/test/java/uk/ac/ebi/atlas/experiments/ExperimentCellCountDaoImplIT.java
alfonsomunozpomer: As long as we use [the same scripts to create fixtures](https://github.com/ebi-gene-expression-group/db-scxa/tree/develop/fixtures), it will pass after adding more experiments. The test can be generalised to check that returns a number greater than 0, but having a value close to 50 means our fixtures are sane and useful.

docker/prepare-dev-environment/postgres/docker-compose.yml
alfonsomunozpomer: Removed!

docker/docker-compose-build.yml
alfonsomunozpomer: Given that it would look like:
```yaml
command:
  - sh
  - -c
  - |
    npm install -g npm-check-updates && \
    ./compile-front-end-packages.sh -ui
```

I thought the version in the code is clearer and consistent with the rest of the services in `docker-compose-build.yml`.

alfonsomunozpomer:
@ke4 I addressed or otherwise clarified the issues you raised. Feel free to resolve the conversations or comment further.

docker/docker-compose-build.yml
ke4: Yes, I reconsidered it and I do agree with you. Thanks for your time

docker/prepare-dev-environment/solr/Dockerfile
ke4: OK, maybe we should have this command in `scxa` as well, but I guess it is OK for now.

ke4 (dismissed):
LGTM 👍

alfonsomunozpomer:
@ke4 There’s bin an extra bit of drama related to how `--env-file` isn’t exactly the same as `env_file` or using `.env` by default. Docker Compose is kind of weird... Have one more look at the last commit, please.

ke4 (approved):
We discussed this in person. It looks good to me. Thanks 👍

PR #348, opened 2023-06-01
Chore – Marker genes only search (again)

alfonsomunozpomer:
I had already merged this PR when I realised that we hadn’t yet merged the related one in `atlas-web-core`, and to enable the test we discussed changes had to happen (in both repos) because `DataFileHub` should be an interface with two implementations rather than having integration tests in the `app` subproject.

Sorry about that!

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -117,7 +117,7 @@ pipeline {
 
         stage('–– Build ––') {
           when { anyOf {
-            branch 'develop'; branch 'main'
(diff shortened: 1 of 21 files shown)


alfonsomunozpomer:
@ke4 I addressed or otherwise clarified the issues you raised. Feel free to resolve the conversations or comment further.

ke4 (dismissed):
LGTM 👍

PR #349, opened 2023-06-07
Fix tsne plot default parameter props

lingyun1010:
In the latest Postgres DB, we have `t-SNE` and `UMAP` rather than `tsne` and `umap`, it should be fixed in the JS bundle, as well as the version of the front-end component `scxa-tsne-plot`.

Please review the front-end component PR https://github.com/ebi-gene-expression-group/atlas-components/pull/140, and I will update the npm package version.


upendrakumbham (dismissed):
All good! Tested this PR on my local machine, working fine with the front-end PR.

upendrakumbham (approved):
ALGTM!

Issue #331, opened 2023-03-03
Number of assays column seems to be always empty in the gene search result table


ke4:
Step to reproduce:

1. Go to SCXA home page and click on `CFTR` link from the list of examples

**Expected result:**

In the resulting table the `Number of assays` should be populated of the number of cells in the experiment.

**Actual result:**

In the resulting table the `Number of assays` missing the number of cells in the experiment.

ke4:
Here is a screenshot of the bug:



![Image](https://user-images.githubusercontent.com/2183410/222786955-d62651e2-014e-4fe0-936c-dfb4a6996733.png)

upendrakumbham:
I verified the backend endpoint, there is no issue with the backend. I could see 'No of Assays" in the payload.

Endpoint : `http://localhost:8080/gxa/sc/json/search?q=RNASEK-C17orf49&species=homo+sapiens`


```
{
   "matchingGeneId":"(ENSG00000161939)",
   "results":[
      {
         "element":{
            "longDescription":"As organisms age, cells accumulate genetic and epigenetic changes that eventually lead to impaired organ function or catastrophic failure such as cancer. Here we describe a single-cell transcriptome analysis of 2544 human pancreas cells from donors, spanning six decades of life. We find that islet cells from older donors have increased levels of disorder as measured both by noise in the transcriptome and by the number of cells which display inappropriate hormone expression, revealing a transcriptional instability associated with aging. By analyzing the spectrum of somatic mutations in single cells from previously-healthy donors, we find a specific age-dependent mutational signature characterized by C to A and C to G transversions, indicators of oxidative stress, which is absent in single cells from human brain tissue or in a tumor cell line. Cells carrying a high load of such mutations also express higher levels of stress and senescence markers, including FOS, JUN, and the cytoplasmic superoxide dismutase SOD1, markers previously linked to pancreatic diseases with substantial age-dependent risk, such as type 2 diabetes mellitus and adenocarcinoma. Thus, our single-cell approach unveils gene expression changes and somatic mutations acquired in aging human tissue, and identifies molecular pathways induced by these genetic changes that could influence human disease. Also, our results demonstrate the feasibility of using single-cell RNA-seq data from primary cells to derive meaningful insights into the genetic processes that operate on aging human tissue and to determine which molecular mechanisms are coordinated with these processes. Examination of single cells from primary human pancreas tissue",
            "experimentDescription":"Single cell transcriptome analysis of human pancreas",
[...]

PR #350, opened 2023-06-15
Number of assays column always empty bug fix

upendrakumbham:
Story reference -  [Number of assays column always empty](https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/331)

- Small fix in the javascript file -  `Experiments.js `
- Tested in my local by regenerating js bundles. I could see the number of assays column values in my local application.


alfonsomunozpomer (approved):
LGTM.

PR #351, opened 2023-07-20
Fix the props and func for marker gene heatmap tab

lingyun1010:
Marker gene heatmap page has a bug on the clusters dropdown menu, the selected value is not showing properly like in the screenshot below. I fixed the bug by testing the in the JS bundle with webpack server, i.e. without in the scxa server. The reviewer can also review it in the same way.

### DIFF
--- a/app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
+++ b/app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
@@ -166,7 +166,7 @@ class TSnePlotViewRoute extends React.Component {
           ks={ks}
           selectedClusterByCategory={search.cellGroupType || search.k || preferredK}
           selectedK={this.state.selectedClusterId}
-          onSelectK={(colourByValue,colourByCategory) => {
+          onSelectK={(colourByValue) => {
               this.setState({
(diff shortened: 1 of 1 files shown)


app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
```diff
@@ -166,16 +165,14 @@ class TSnePlotViewRoute extends React.Component {
           wrapperClassName={`row expanded`}
           ks={ks}
           selectedClusterByCategory={search.cellGroupType || search.k || preferredK}
-          selectedClusterId={this.state.selectedClusterId}
-          onChangeClusterId={(colourByCategory, colourByValue) => {
+          selectedK={this.state.selectedClusterId}
+          onSelectK={(colourByValue,colourByCategory) => {
```
alfonsomunozpomer: The parameter `colourByCategory` isn’t passed and it’s also not in the function signature. Let’s remove it. I guess it’s a leftover from the previous `onChangeClusterId`.

alfonsomunozpomer (changes requested):
One minor change and good to go.

alfonsomunozpomer (approved):
LGTM.

PR #352, opened 2023-07-31
Refactor organism part endpoint

ke4:
Organism parts endpoint need to process query params: species, cell types and is marker gene parameters.


app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -130,7 +131,7 @@ public String search(@RequestParam MultiValueMap<String, String> requestParams)
         // If the query term matches a single Ensembl ID different to the query term, we return it in the response.
         // The most common case is a non-Ensembl gene identifier (e.g. Entrez, MGI, ...).
         var matchingGeneIds =
-                (geneIds.get().size() == 1 && !geneIds.get().iterator().next().equals(geneQuery.queryTerm())) ?
+                (Objects.requireNonNull(geneIds.orElse(null)).size() == 1 && !geneIds.get().iterator().next().equals(geneQuery.queryTerm())) ?
```
upendrakumbham: I'm looking for more clarification on this as I'm confused here.`(Objects.requireNonNull(geneIds.orElse(null)).size() == 1`

upendrakumbham: I mean in my opinion, this needs to be simplified like below:

`this.bar = Objects.requireNonNull(bar, "bar must not be null");`

or 

with static import, we can simplify further:

`this.bar = requireNonNull(bar, "bar must not be null");`

ke4: @upendrakumbham I think you misread the code.
It is not `Objects.requireNonNull(geneIds).orElse(null)` but `Objects.requireNonNull(geneIds.orElse(null))`
Could you check it please once again?
Or we can discus it later today?
Thanks

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -249,7 +251,10 @@ public Set<String> getOrganismPartBySearchTerm(@RequestParam MultiValueMap<Strin
             return ImmutableSet.of();
         }
 
-        return analyticsSearchService.searchOrganismPart(geneIds.get());
+        var cellTypes = requestParams.get("cellTypes");
```
upendrakumbham, line 254:
```
if (geneIds.isEmpty()) {
            return ImmutableSet.of();
        }

        var cellTypes = requestParams.get("cellTypes");
return organismPartSearchService.search(geneIds.get(),
                cellTypes != null ? ImmutableSet.copyOf(cellTypes) : ImmutableSet.of());


```

Instead of the above, how about this code: 

```suggestion
if (!geneIds.isEmpty() && equestParams.get("cellTypes") !-null) {
     return organismPartSearchService.search(geneIds.get(), 
                ImmutableSet.copyOf(cellTypes));
 }
 else {
    return ImmutableSet.of()
 }
```

Please check the Indentation of my code in your IDE if you want to take it :). 
If you think this review is not related to endpoint improvement, ignore this.

upendrakumbham: sorry, my bad, correct this copy and paste error in the above suggestion - `equestParams.get("cellTypes") !-null` should be `requestParams.get("cellTypes") !=null`

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: Ok, I found a much better way to write this code fragments.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: @upendrakumbham I don't really understand your comment. I don't see why is that better.
Actually that code modification that you are suggestion would break the current logic.
Have you followed the code path that you would like me to modify?
If not, then please do. Thanks

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -236,7 +238,7 @@ public Boolean isMarkerGene(@RequestParam MultiValueMap<String, String> requestP
                                 .map(Map.Entry::getKey)
                                 .collect(toImmutableSet()));
 
-        return markerGeneFacets != null && markerGeneFacets.size() > 0;
+        return markerGeneFacets != null && !markerGeneFacets.isEmpty();
```
upendrakumbham, line 241: We can use `StringUtils.isBlank()`-> this would check white spaces and null.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
upendrakumbham: Sorry, my bad. I mixed up. In my opinion, Positive checks come first(! == something)->{do some work} then negative check last if(empty)-->return empty.

upendrakumbham: Sorry, my bad. I mixed it up.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: @upendrakumbham `markerGeneFacets` is not a `String`. Its type is: `ImmutableMap<String, Map<String, Map<Integer, List<Integer>>>>`. Could you tell me please where would you like to use `StringUtils.isBlank()`?
What I am checking here is: are there any marker gene based on the given request parameters.

upendrakumbham (approved):
All looks good to me. Please merge away!

PR #353, opened 2023-08-01
Refactor cell types endpoint

### LINKED ISSUE #356: Filter out null element from the resulted list from cell types endpoint
Currently if we do a query like this: http://localhost:8080/gxa/sc/json/gene-search/cell-types?symbol=CFTR
then there is a `null` element in the returned result.
We need to filter that out.

ke4:
Cell types endpoint need to process query params: organism parts, species and is marker gene.


app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -262,7 +267,10 @@ public Set<String> getCellTypeBySearchTerm(@RequestParam MultiValueMap<String, S
             return ImmutableSet.of();
         }
 
-        return analyticsSearchService.searchCellType(geneIds.get());
+        var organismParts = requestParams.get("organismParts");
```
upendrakumbham, line 270: I think the organism part endpoint code refactor is applicable here too. Code implementation is similar.

Will start with a positive condition first instead of a negative check.

app/src/main/java/uk/ac/ebi/atlas/search/celltype/CellTypeSearchService.java
```diff
@@ -0,0 +1,29 @@
...
+@RequiredArgsConstructor
+public class CellTypeSearchService {
+
+    private final CellTypeSearchDao cellTypeSearchDao;
+    private final GeneSearchService geneSearchService;
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(CellTypeSearchService.class);
+
+    public ImmutableSet<String> search(ImmutableSet<String> geneIds, ImmutableSet<String> organismParts) {
+        if (geneIds.isEmpty()) {
```
upendrakumbham, line 20: We can add not empty if condition first and then empty one please.
```
suggestion

if (!geneIds.isEmpty()) {
LOGGER.info("Searching organism parts for this gene ids: {}", geneIds.asList());
  return cellTypeSearchDao.searchCellTypes(geneSearchService.getCellIdsFromGeneIds(geneIds), organismParts);

}
else
{
   LOGGER.warn("Can't query for organism part as no gene IDs has given.");
            return ImmutableSet.of();
}
```

upendrakumbham (changes requested):
Minor code improvements are required to merge, the rest of the code is good.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
ke4: I am sorry, but your comment is not clear to me.
Could you tell me with a bit more context what you mean, please?

app/src/main/java/uk/ac/ebi/atlas/search/celltype/CellTypeSearchService.java
ke4: @upendrakumbham These articles hopefully help you understand why fail-fast is a better way to write your code:
- https://www.martinfowler.com/ieeeSoftware/failFast.pdf
- https://en.wikipedia.org/wiki/Fail-fast
- https://softwareengineering.stackexchange.com/questions/254002/checking-negative-of-a-condition
Please let me know if it is not clear to you. We can discuss this F2F, if you prefer...
Thanks

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
```diff
@@ -236,7 +238,7 @@ public Boolean isMarkerGene(@RequestParam MultiValueMap<String, String> requestP
                                 .map(Map.Entry::getKey)
                                 .collect(toImmutableSet()));
 
-        return markerGeneFacets != null && markerGeneFacets.size() > 0;
+        return markerGeneFacets != null && !markerGeneFacets.isEmpty();
```
upendrakumbham: We can use one statement `StringUtils.isBlank()`-> this would check white spaces and null.

app/src/main/java/uk/ac/ebi/atlas/search/JsonGeneSearchController.java
upendrakumbham: I have gone through the all articles you mentioned. I agree with you. Thanks.

app/src/main/java/uk/ac/ebi/atlas/search/celltype/CellTypeSearchService.java
upendrakumbham: Agree with you. Thanks.

upendrakumbham (dismissed):
All looks good to me. Please merge away!

PR #354, opened 2023-08-30
Improve Docker Compose scripts to prepare the dev environment

alfonsomunozpomer:
Some fixes to the run.sh  scripts and YAML files. They solve some issues you can have when you populate an environment from scratch.


lingyun1010 (approved):
Yes works for me. LGTM.

PR #370, opened 2023-11-17
Replace lodash function with JS function

### LINKED ISSUE #361: Showing gray color plot expression when we click on specific experiment for the first time page load
In the SCXA a few experiments don't have metadata. Those experiments have given empty colour plot expressions when clicking on the experiment.

`Ex: https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-GEOD-99058/results/tsne`

Here is the screenshot:

Image

Change the K value to something and return to the same k value that is `23`, then you will see a colour plot.

-  We are getting an empty colour plot for the metadata available experiments too.

Image

Empty colour plot  screen shot:

Image

lingyun1010:
The tsne plot is not showing properly. Typically it is due to the lodash function is not compiled well for some reasons, so I replaced the function with JS build-in function.


ke4 (approved):
LGTM
You can merge this to `develop` branch, but we won't deploy that version for now.
I am going to cherry-pick this change to the latest hotfix branch and deploy that one.
Thanks

PR #384, opened 2024-01-24
Update SQL for non-priority case in cell plot

### LINKED ISSUE #374: Experiment's default plot option is not correctly displayed in SCXA
We discussed the use case that should be implement for this with @upendrakumbham and @@pmb59.
There are 2 separate condition that we have to fullfill:

- If in the `priority` column in `scxa_dimension_reduction` table there is a **non zero** value (most probably 1) for a specific plot type for a specific experiment then the web app should show that one as default.
- If every value in the `priority` column in `scxa_dimension_reduction` table is **0**, then the web app should calculate a middle value for `Plot options` and display that one. For example: if we have 9 `Plot options` then we should display the 5th one.

Steps to reproduce:
1. Go to `Browse experiments` tab in `SCXA`.
2. Select [Transcriptional programs of circuit assembly in the Drosophila visual system](https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-GEOD-156455/results)

Expected result:
The middle `Plot option` is selected, as all the `priority` column in `scxa_dimension_reduction` table is **0**.

Actual result:
The 1st `Plot option` is selected.

You should also test it with `E-ANND-3` in your local environment. Please load that experiment into your local environment , if you have not done it, yet. The instructions to follow: [Add an AnnData experiment to our local SCXA](https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/340)

Please see below how to reproduce the bug with that experiment.

-----
For historical reason I preserved the original ticket description and you can find it below.

Original title: **`E-ANND-3` experiment by default is not displaying plots in SCXA**

Steps to reproduce:
1. Go to `Browse experiments` tab in `SCXA`.
2. Select [Tabula Sapiens - Smart-seq2 dataset](http://localhost:8080/gxa/sc/experiments/E-ANND-3/results) experiment

**Expected result:**

Experiment's data should show up with the plots diagram.

**Actual result:**

An error message appears on the screen.

Image

lingyun1010:
This a bugfix related to the issue https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/374.

In the previous implementation, we suppose at least one highest `priority` value, i.e. larger than `0`, would be curated in the table `scxa-dimension-reduction`. However, due to lack of enough data or curated by mistakes, we do not always have the highest `prioriy` value, so I fix the postgres query to select proper entries in both cases. 

There are some data populating issues on dev PostgreSQL database, which has been reported to data production team members on Slack. Please follow the updates via https://ebi-fg.slack.com/archives/C7U2CRS58/p1706111909985499.

### DIFF
--- a/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotDao.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotDao.java
@@ -157,14 +157,15 @@ public class CellPlotDao {
 
 
     private static final String SELECT_DEFAULT_PLOT_METHOD_AND_PARAMETERISATION =
-            "SELECT dr.method, jsonb_array_elements(dr.parameterisation) parameterisation " +
+            "SELECT DISTINCT dr.method, jsonb_array_elements(dr.parameterisation) parameterisation " +
                     "FROM scxa_dimension_reduction dr " +
                     "JOIN (SELECT method,  max(priority) as prt " +
                         "FROM scxa_dimension_reduction " +
                         "WHERE experiment_accession=:experiment_accession " +
                         " GROUP BY method) fi " +
                     "ON dr.method = fi.method " +
-                    "AND dr.priority = fi.prt";
+                    "AND dr.priority = fi.prt " +
+            "ORDER BY parameterisation";
 
     public Map<String, List<JsonObject>> fetchDefaultPlotMethodWithParameterisation(String experimentAccession) {
         var namedParameters = ImmutableMap.of("experiment_accession", experimentAccession);
--- a/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotDao.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/experimentpage/cellplot/CellPlotDao.java
@@ -165,6 +165,7 @@ public class CellPlotDao {
                         " GROUP BY method) fi " +
                     "ON dr.method = fi.method " +
                     "AND dr.priority = fi.prt " +
+            "WHERE experiment_accession=:experiment_accession " +
             "ORDER BY parameterisation";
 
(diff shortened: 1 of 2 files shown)


upendrakumbham (changes requested):
Please add a test case for this change or modify it if it already exists.

lingyun1010:
@upendrakumbham reported a bug regarding t-SNE plots. The default `umap` plot parameterisation for experiment `E-ANND-3` is `n_neighbors:20`, however the only entry in the postgres table is `n_neighbors:15`.

Due to I don't have `E-ANND-3` experiment data yet locally, to debug the issue, I temporarily deleted the table entries for the experiment `E-EHCA-2` for `umap` plot methods, including `n_neighbors:10,100,20,25,3,30,5,50` and only left `n_neighbors:15`.

I can generate the same bug locally in this way. To check the default parameterisation value I check the json endpoint `http://localhost:8080/gxa/sc/json/cell-plots/E-EHCA-2//default/plot-method` and I got 
```
{"tsne":{"perplexity":25},"umap":{"n_neighbors":20}}
```

In conclusion by now, the bug comes from the backend not frontend issue and we may encounter the same bug if the database has only one entry for one of the cell plot type method.

lingyun1010:
I tested the relevant SQL implemented in `CellPlotDao.fetchDefaultPlotMethodWithParameterisation` as following: 
```
SELECT DISTINCT dr.method, jsonb_array_elements(dr.parameterisation) parameterisation
                    FROM scxa_dimension_reduction dr
                    JOIN (SELECT method,  max(priority) as prt
                        FROM scxa_dimension_reduction
                        WHERE experiment_accession=:experiment_accession
                         GROUP BY method) fi
                    ON dr.method = fi.method
                    AND dr.priority = fi.prt
            ORDER BY parameterisation
```
where `experiment_accession` = `E-EHCA-2` as my test case.

I get a wrong query response, which contains all the other `n_neighbors` options. Hence, the bug is from the above SQL.

upendrakumbham (dismissed):
All looks good, please fix the build before merge.

ke4 (approved):
LGTM

PR #388, opened 2024-03-21
Add consistency between cell plots tab and marker genes tab

### LINKED ISSUE #381: Inconsistency between cell plot and heat map
The number of clusters isn’t conserved between the cell plot and marker gene tabs. There are two request parameters, one for each component (k and colourBy, respectively): experiments/{experiment_accession}/results/marker-genes?k=19&colourBy=8

Steps to reproduce:
1. Go to `Browse experiments` page
2. Select [A lymphocyte-microglia-astrocyte axis in chronic active multiple sclerosis](https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-GEOD-180759/results)
3. Select `k=13` from the `Colour plot by` dropdown. URL is going to be: https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-GEOD-180759/results?colourBy=13
4. Click on `Marker Genes` tab

**Expected result:**
URL should be: https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-GEOD-180759/results/marker-genes?colourBy=13&k=38 (default cluster k value)

**Actual result:**
URL is: https://wwwdev.ebi.ac.uk/gxa/sc/experiments/E-GEOD-180759/results/marker-genes?colourBy=13

lingyun1010:
According to Silvie's detailed comments, I add logics to make cell plot & marker gene plot options are consistent when the plot type is `clusters`.

Besides, I refactored the tab name from `tsne` to `cell-plots` as discussed in some meetings before, so the tab refers to both tsne and umap plots more properly.

### DIFF
--- a/app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
+++ b/app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
@@ -29,7 +29,7 @@ RedirectWithSearchAndHash.propTypes = {
 const RedirectWithLocation = withRouter(RedirectWithSearchAndHash)
 
 const CLUSTERS_PLOT = `clusters`
-const METADATA_PLOT = `metatdata`
+const METADATA_PLOT = `metadata`
 
 class TSnePlotViewRoute extends React.Component {
   constructor(props) {
--- a/app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
+++ b/app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
@@ -49,12 +49,12 @@ class TSnePlotViewRoute extends React.Component {
     const { species, experimentAccession, accessKey, ks, ksWithMarkerGenes, plotTypesAndOptions, metadata, anatomogram } = this.props
     const search = URI(location.search).search(true)
 
-let plotTypeDropdown =
-        Object.keys(defaultPlotMethodAndParameterisation)
-            .map(plot => ({
-              plotType: plot,
-              plotOptions: plotTypesAndOptions[plot]
-            }));
+    let plotTypeDropdown =
+      Object.keys(defaultPlotMethodAndParameterisation)
+        .map(plot => ({
+          plotType: plot,
+          plotOptions: plotTypesAndOptions[plot]
+        }))
 
     let organWithMostOntologies = Object.keys(anatomogram)[0]
     for (let availableOrgan in anatomogram) {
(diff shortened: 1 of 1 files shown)


app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
```diff
@@ -28,6 +28,9 @@ RedirectWithSearchAndHash.propTypes = {
 
 const RedirectWithLocation = withRouter(RedirectWithSearchAndHash)
 
+const CLUSTERS_PLOT = `clusters`
+const METADATA_PLOT = `metatdata`
```
ke4: I guess it is a typo.
```suggestion
const METADATA_PLOT = `metadata`
```

app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
```diff
@@ -74,7 +77,7 @@ class TSnePlotViewRoute extends React.Component {
 
     const routes = [
       {
-        path: `/tsne`,
+        path: `/cell-plots`,
```
ke4, line 68: This is a good change that I suggested a long time ago. Thanks for this one!

ke4 (changes requested):
There is a tiny typo that I commented.
My bigger concern is that this JS file is not very readable.
I would like you to try to refactor this file using the knowledge you learnt either on the Uncle Bob or Venkat Subramaniam videos.
Many thanks,
Karoly

ke4:
Your latest modification improved the code a bit, but we could do much more with it.
I was thinking on this a bit, and I TBH it is going to be easier if I am going to sit down with you and show you how to clean this code. We can do it tomorrow if you are going to be in the Campus. It will only take a couple of minutes to show you and after that probably 10-15 minutes of your time to make the code prettier.
Thanks

app/src/main/javascript/bundles/experiment-page/src/TSnePlotViewRoute.js
```diff
@@ -42,28 +37,24 @@ class TSnePlotViewRoute extends React.Component {
...
-        plotType: Object.keys(defaultPlotMethodAndParameterisation)[1],
-        plotOptions: plotTypesAndOptions[Object.keys(defaultPlotMethodAndParameterisation)[1]]
-      }
-    ]
+    let plotTypeDropdown =  []
+    Object.keys(defaultPlotMethodAndParameterisation).map(plot => plotTypeDropdown.push(
+    {
+      plotType: plot,
+          plotOptions: plotTypesAndOptions[plot]
+    }))
```
ke4: Eliminates the creation an empty object first and use of the `push` method later. `map` is already doing all of that.
```suggestion
let plotTypeDropdown =
        Object.keys(defaultPlotMethodAndParameterisation)
            .map(plot => ({
              plotType: plot,
              plotOptions: plotTypesAndOptions[plot]
            }));
```

ke4 (changes requested):
I added a suggestion that you can quickly apply with the use of the button.
I think there are other things we could improve, but I don't want to spend more time on this one, so I am going to approve it after you applied my suggestion.
Thanks

ke4 (approved):
LGTM!

PR #397, opened 2024-05-03
[Feature] Improve Anndata experiment page appearance

### LINKED ISSUE #393: Modify the title session of an Anndata experiment
If the experiment is an Anndata experiment, then we have to differentiate it from other experiments.
In the title session we have to do the following changes:
- have a different shade in the background (specify the colour with @Silvie)
- add the Anndata specific icon to the right side of the title session

Image

ke4:
If the displayed experiment is an `Anndata` experiment then we use a different background colour for the experiment description section and add its icon to the right side of this section.


lingyun1010 (changes requested):
Looks good! I suggest to add a tooltip for anndata icon for users to provide some hints that it stands for `anndata` in our website.

lingyun1010 (approved):
LGTM! Please update to the latest develop before merging.

ke4:
> LGTM! Please update to the latest develop before merging.

There is no conflict with the latest develop. I already updated this branch with develop in the morning.

PR #414, opened 2024-05-30
[Chore] Add no build flag to local deployment

ke4:
Add the `-n` parameter to the `build-and-deploy-webapp.sh` shell script to be able to startup the application without UI and/or backend build. We already have this option in `bulk` I just ported it to `single cell` web app.


upendrakumbham (approved):
All looks good to me!

PR #428, opened 2024-07-01
[Bugfix] Fix processing search data for cell type wheel

### LINKED ISSUE #424: Cell type wheel visualise the data wrongly
Cell type search result page not working correctly. No gene heat map showing on the right, after clicking on any cell in the left wheel.

Cell Type wheel in `staging` environment (not correct):

Image

Cell type wheel in `public` environment (correct):

Image

Steps to reproduce:

1. Open Single Cell Expression Atlas on `staging` environment: https://wwwdev.ebi.ac.uk/gxa/sc/home
2. Select `liver (organ/organism part)` on the gene search from the home page: https://wwwdev.ebi.ac.uk/gxa/sc/search/metadata/liver

ke4:
The search result processing for the cell type wheel was completely broken.
This change fixing the following issues:
- remove an element from the search result if the element's key equals with the search term. It happens when we search for a species.
- returns an empty `ImmutableSet` if the search result is empty. Before this change it returned only the search term and that was visualised in the cell type wheel. 
This change also completely replaces the wrongly written test case and adds more test cases to cover each cases.

### DIFF
--- a/app/src/main/java/uk/ac/ebi/atlas/search/metadata/HighchartsSunburstAdapter.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/search/metadata/HighchartsSunburstAdapter.java
@@ -7,7 +7,7 @@ import com.google.common.collect.ImmutableSet;
 import org.apache.commons.lang3.tuple.ImmutablePair;
 import org.springframework.stereotype.Component;
 
-import java.util.UUID;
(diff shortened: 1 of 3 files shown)


upendrakumbham (approved):
All looks good to me.  I have gone through test case scenarios and good coverage.

PR #435, opened 2024-07-18
[bugfix] Update scxa-gene-search-form and scxa-cell-type-wheel-heatmap to latest

ke4:
For passing the `species` parameter from the web app's homepage from the gene search feature to the cell type wheel component we had to update our `scxa-gene-search-form` and `scxa-cell-type-wheel-heatmap` UI component to their latest version where we fixed this issue.


lingyun1010 (approved):
LGTM!

PR #440, opened 2024-07-25
[bugfix] Render big dataset with Highcharts in the Single Cell web app

ke4:
To be able to render big datasets on the cell type wheel we updated the `scxa-cell-type-wheel-heatmap` component in the `atlas-component` repo and we are using that latest/updated version of that component in this web application.
I aso updated the `faceted search result` component to the latest version.


lingyun1010 (approved):
LGTM！

PR #442, opened 2024-07-30
Filter species metadata cell type wheel search on ogranism field

### LINKED ISSUE #439: Filter species metadata cell type wheel search on ogranism field
In the metadata search result page, if the search term is a species, such as `Homo sapiens` or `Mus musculus`, some entry in the json response is from other species. Given some results in Solr, for example, `Homo sapiens`, 

```
(ontology_annotation_ancestors_labels_t:Homo sapiens OR ontology_annotation_parent_labels_t:Homo sapiens OR ontology_annotation_part_of_rel_labels_t:Homo sapiens OR ontology_annotation_synonyms_t:Homo sapiens OR ontology_annotation_label_t:Homo sapiens) AND experiment_accession:E-GEOD-156455
```
Image

there is an entry contains `Homo sapiens` in one of the ontology labels, but from `Drosophila melanogaster`.

To prevent this error happens, we need to filter the search results on `ctw_organism` if it is a **species** search.
i.e. `ctw_organism==species_search_term`.

lingyun1010:
In the cell type wheel json response with `species` search term, for example, `Mus musculus` or `Homo sapiens`, due to the solr contains some cross-species ontology data, the results have other species. To resolve this problem, we need to create a filter on searching if the search term is a species.

Rather than passing a request parameter like, I think it's better to check if it's a species search in the back-end, as we don't have to call the `allSpecies` endpoints from front-end.

### DIFF
--- a/app/src/main/java/uk/ac/ebi/atlas/search/metadata/CellTypeWheelDao.java
+++ b/app/src/main/java/uk/ac/ebi/atlas/search/metadata/CellTypeWheelDao.java
@@ -81,11 +81,37 @@ public class CellTypeWheelDao {
         singleCellAnalyticsCollectionProxy = proxyFactory.create(SingleCellAnalyticsCollectionProxy.class);
     }
 
-    public ImmutableList<ImmutableList<String>> facetSearchCtwFields(String searchTerm, String species, boolean isSpeciesSearch) {
-        // Build the facet in the query above:
-        // "facet": {
-        //   ...
-        // }
+    public ImmutableList<ImmutableList<String>> facetSearchCtwFields(String searchTerm, String species) {
+        var queryBuilder = buildSolrQueryForCellTypeWheel(searchTerm);
+        // If there’s a species add it as an additional filter
+        if (isNotBlank(species)) {
+            queryBuilder.addFilterFieldByTerm(CTW_ORGANISM, species);
+        }
+
+        return extractNestedFacetTerms(
+                (SimpleOrderedMap<Object>) singleCellAnalyticsCollectionProxy
+                        .query(queryBuilder)
+                        .getResponse()
+                        .findRecursive("facets"),
+                ImmutableList.of(ORGANISMS_TERMS_KEY, ORGANISM_PARTS_TERMS_KEY, CELL_TYPES_TERMS_KEY, EXPERIMENT_ACCESSIONS_TERMS_KEY),
+                ImmutableList.of());
+    }
+
+    public ImmutableList<ImmutableList<String>> speciesSearchCtwFields(String searchTerm, String species) {
+        var queryBuilder = buildSolrQueryForCellTypeWheel(searchTerm);
+
+        // If the search term is a species add it as an additional filter
+        queryBuilder.addFilterFieldByTerm(CTW_ORGANISM, searchTerm);
+
+        return extractNestedFacetTerms(
+                (SimpleOrderedMap<Object>) singleCellAnalyticsCollectionProxy
+                        .query(queryBuilder)
+                        .getResponse()
+                        .findRecursive("facets"),
+                ImmutableList.of(ORGANISMS_TERMS_KEY, ORGANISM_PARTS_TERMS_KEY, CELL_TYPES_TERMS_KEY, EXPERIMENT_ACCESSIONS_TERMS_KEY),
+                ImmutableList.of());
+    }
+    private SolrQueryBuilder<SingleCellAnalyticsCollectionProxy> buildSolrQueryForCellTypeWheel(String searchTerm) {
         var facetBuilder = new SolrJsonFacetBuilder<SingleCellAnalyticsCollectionProxy>()
                 .setFacetField(CTW_ORGANISM)
                 .addNestedFacet(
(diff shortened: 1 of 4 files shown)


app/src/main/java/uk/ac/ebi/atlas/search/metadata/CellTypeWheelService.java
```diff
@@ -13,13 +14,19 @@
...
+    public CellTypeWheelService(CellTypeWheelDao cellTypeWheelDao,
+                                FeaturedSpeciesService featuredSpeciesService) {
         this.cellTypeWheelDao = cellTypeWheelDao;
+        this.featuredSpeciesService = featuredSpeciesService;
     }
 
     public ImmutableSet<ImmutablePair<ImmutableList<String>, String>> search(String searchTerm, String species) {
-        return cellTypeWheelDao.facetSearchCtwFields(searchTerm, species)
+        ImmutableList<String> allSpeciesNames = featuredSpeciesService.getSpeciesNamesSortedByExperimentCount();
+        var isSpeciesSearch = allSpeciesNames.contains(searchTerm);
```
ke4: As `featuredSpeciesService.getSpeciesNamesSortedByExperimentCount()` call is returning a capitalised list of species name we also need to capitalise the `searchTerm` variable before we call `contains()` method on the list of species.

```suggestion
        var isSpeciesSearch = allSpeciesNames.contains(StringUtils.capitalize(searchTerm));
```

app/src/test/java/uk/ac/ebi/atlas/search/metadata/CellTypeWheelDaoIT.java
ke4, line 1: I could not find any tests that passing `true` to the new parameter (`isSpeciesSearch`) to this modified method:
`facetSearchCtwFields` of the `CellTypeWheelDao` class.
We should add those test cases, too.
Thanks

app/src/test/java/uk/ac/ebi/atlas/search/metadata/CellTypeWheelServiceTest.java
ke4, line 1: Same missing as in `CellTypeWheeldaoIT`:
I could not find any tests that passing true to the new parameter (isSpeciesSearch) to this modified method:
`facetSearchCtwFields` of the `CellTypeWheelDao` class.
We should add those test cases, too.
Thanks

ke4 (changes requested):
@lingyun1010 
Thanks for your work. Good work, but I added some comments that I would like you to consider.
Thanks

app/src/main/java/uk/ac/ebi/atlas/search/metadata/CellTypeWheelService.java
lingyun1010: Yes, and our species search term is also capitalized, so we don't need to add this utils anymore.

ke4: Yes, for now on front end side we have this line in `LabelledSelect.js` to capitalise the species name:
```
const capitalisedName = name.charAt(0).toUpperCase() + name.slice(1)
```
But for future proofing, - as the code can change in the future - I would like you to add capitalisation in the back end side, too.
I think, it is very similar when you validate input parameters both on UI and later on back-end side as anybody can change any input parameter on frontend side with a simple tool.
It is a good practise, I believe.
Thanks

app/src/main/java/uk/ac/ebi/atlas/search/metadata/CellTypeWheelDao.java
```diff
@@ -81,7 +81,7 @@ public CellTypeWheelDao(SolrCloudCollectionProxyFactory proxyFactory) {
         singleCellAnalyticsCollectionProxy = proxyFactory.create(SingleCellAnalyticsCollectionProxy.class);
     }
 
-    public ImmutableList<ImmutableList<String>> facetSearchCtwFields(String searchTerm, String species) {
+    public ImmutableList<ImmutableList<String>> facetSearchCtwFields(String searchTerm, String species, boolean isSpeciesSearch) {
```
ke4: We had this case in one of the Uncle Bob videos and probably also in the Venkat Subramaniam training, as well.
The code would be much cleaner and more understandable if we create 2 well named methods from this method without the boolean parameter.
That would also help you clean the above conditions in line 118 and 123.
Thanks

ke4 (changes requested):
I added 2 change requests.
Could you please apply them?
Thanks

app/src/main/java/uk/ac/ebi/atlas/search/metadata/CellTypeWheelDao.java
```diff
@@ -82,10 +82,39 @@ public CellTypeWheelDao(SolrCloudCollectionProxyFactory proxyFactory) {
     }
 
     public ImmutableList<ImmutableList<String>> facetSearchCtwFields(String searchTerm, String species) {
-        // Build the facet in the query above:
-        // "facet": {
-        //   ...
-        // }
+        var queryBuilder = buildSolrQueryForCellTypeWheel(searchTerm);
```
ke4: There are couple of duplications in these 2 methods (`facetSearchCtwFields` and `speciesSearchCtwFields`).
Could you please extract them to private methods?
Thanks

app/src/main/java/uk/ac/ebi/atlas/search/metadata/CellTypeWheelService.java
```diff
@@ -13,19 +15,32 @@
...
+        return isSpeciesSearch ? cellTypeWheelDao.speciesSearchCtwFields(searchTerm, species)
                 .stream()
                 // This will effectively “explode” tuples and aggregate experiment accessions (the last element in the
                 // tuple) to the organisms, organism parts and cell types
                 .map(this::addTailToEveryHeadSublist)
                 .flatMap(ImmutableList::stream)
-                .collect(toImmutableSet());
+                .collect(toImmutableSet()) :
+                cellTypeWheelDao.facetSearchCtwFields(searchTerm, species)
+                        .stream()
```
ke4: There is code duplication here. You can extract the results from the DAO to a variable and do the streaming on that.

ke4 (changes requested):
I added 2 comments about DRYing the code.
Could you apply them, please?
Thanks

ke4 (approved):
LGTM!
Thanks for all your patience for applying all my comments/suggestions. ;-)

PR #446, opened 2024-08-06
Bugfix search button not working after an initial metadata search

### LINKED ISSUE #431: Search button not working after an initial metadata search
Steps to reproduce:

1. Go to metadata search and search for 'liver`
2. Select species: `Callithrix jacchus` and search term: `LIAS` (from Symbol list)
3. Click on the `Search` button
4. Select species: `Homo sapiens` and search term: `liver` (from Metadata list)
5. Click on the `Search` button

Expected behaviour:
You get a result for the selected search term filtered by the selected `species`.

Actual behaviour:
Search is not triggered.

lingyun1010:
The bug is when we search a gene and we will land in gene search result page, the endpoint is different with the metadata search, so the metadata search will not be triggered. The solution we discussed is to correct the router/url for metadata search.


ke4 (approved):
LGTM!

Issue #447, opened 2024-08-07
Make experiment unit adaptable not hard-coded in marker gene heatmap


lingyun1010:
In Single Cell web app the units that appears in the `Marker Genes` tab of the `Experiment page` below the title: `Cluster marker genes` looks like they are hardcoded.
They are displaying: `Median expression (CPM)`

We should get the units from our Postgres DB. We are storing it in the `experiment` table's `expression_unit` field.

ke4:
It is hardcoded in 5 different places.

1. `scxa-marker-gene-heatmap` component:
In the `MarkerGeneHeatmap.js` we need to change this lines to get the units from an input variable / property.

in line, 245 in the current version of the file:
```
    legend: {
      title: {
        text: `Median expression (CPM)`
      },
      align: `center`,
      verticalAlign: `top`,
      layout: `horizontal`,
      symbolWidth: 480,
      enabled: data.length !== 0
    },
```

2. `scxa-marker-gene-heatmap` component:
in `heatmapOptionsProvider.js`, line 16 in the current version of the file:
```
<b>Expression:</b> ${+this.point.value.toFixed(3)} CPM
```

3. `scxa-marker-gene-heatmap` component:
in `heatmapOptionsProvider.js`, line 46 in the current version of the file:
```
<b>Expression:</b> ${+this.point.value.toFixed(3)} CPM
```

4. `scxa-marker-gene-heatmap` component:
in `heatmapOptionsProvider.js`, line 67 in the current version of the file:
```
<b>Median expression:</b> ${+this.point.value.toFixed(3)} CPM
```

5. `scxa-tsne-plot` component:
in `GeneExpressionTSnePlot.js`, line 167 in the current version of the file:
```
{
      title: {
        text: "Expression level (CPM)"
      },
      floating: false,
      align: "center",
      symbolHeight: 5,
      symbolWidth: 450,
      enabled: true
    }
```

ke4:
And there is a 6th hardcoded, but that is coming from the backend: `CellPlotJsonSerializer.java`, line 91:
```
        var unit = "CPM"; // Get units from experiment, or from request parameter if more than one is available
```
This should be changed and getting that value from the DB.
This is a must have fix for the current release.

ke4:
PR for the 6th case: [[Chore] Get expression unit from the DB for Cell plots tab / Gene Expression part #449](https://github.com/ebi-gene-expression-group/atlas-web-single-cell/pull/449)

PR #449, opened 2024-08-08
[Chore] Get expression unit from the DB for Cell plots tab / Gene Expression part

ke4:
Currently the `expression unit` of an experiment is hardcoded everywhere to `CPM`.
This change is going to fix it in the cell plot diagram (in the tooltip).
We are already storing the value of the expression unit in the DB, in the `experiment` table.
We are going to get it from there and pass it to the UI.

I tested it locally and it work fine. 

This is fixing the 6th case in this ticket: [Make experiment unit adaptable not hard-coded in marker gene heatmap](https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/447)	https://github.com/ebi-gene-expression-group/atlas-web-single-cell/issues/447
According to Silvie this was the highest priority.
We need to create other PRs for the other cases.


lingyun1010 (dismissed):
LGTM!

ke4:
@lingyun1010 is already approved this PR, but after her approval I accidentally pushed another commit to this PR as it was a very similar case, but I realised that it is going to need another PR, so I removed that commit from this PR.
I am going to force `Squash and merge` this PR now.
Sorry about this.

PR #450, opened 2024-08-09
Experiment unit from db for marker genes

### LINKED ISSUE #447: Make experiment unit adaptable not hard-coded in marker gene heatmap

ke4:
This change add expression unit to the MarkerGene entity to provided that data for the front end component to be able to display it dynamically, rather than by a hardcoded value.

WIP


upendrakumbham (approved):
LGTM!

PR #451, opened 2024-08-14
Update tsne plot package version and ks options

### LINKED ISSUE #163: Cell plots ks are not always required for anndata experiments
Experiment page for `E-ANND-3` has errors ` this.props.ks[0] is undefined`: https://www-test.ebi.ac.uk/gxa/sc/experiments/E-ANND-3/results/.

Given the information from @irisdianauy , typical Atlas experiments will always have ks but will not always have inferred cell types if I understand correctly.  But for ANND experiments, it’s the opposite: they will always have inferred cell types, but may not always have ks.

lingyun1010:
As we don't have any data locally without `ks` for now, I just published a war file to test environment to test. For `E-ANND-3` the ks will be empty array. 

In addition, our local database is not fully update for this experiment, and we properly need to do that after the scxa release.


upendrakumbham (approved):
LGTM!

PR #457, opened 2024-09-02
Add more experiment to anatomogram black list

### LINKED ISSUE #453: Update Anatomogram Black list  Experiments in the SCXA
We will receive blacklisted experiments from the @silvie or data prod team.

That list , we need to update here as we have already added here - `ExperimentPageContentService` class

lingyun1010:


ke4 (approved):
LGTM!

PR #458, opened 2024-09-03
[Feature] Add natural sorting order to clustered marker genes for visualising them in our Heatmap component

### LINKED ISSUE #456: Make cell plots k options compatible with unsupervised cluster ids
We need to refactor the handling / creating of cluster names.
Currently it looks like it is restricted to use only numerical names, like `Cluster 1`, `Cluster 2` and so on.
With the introduction of external datasets (`anndata`) that we are not re-analysing we could get strings and not numbers as cluster names. That means we need to able to handle any strings as the name of the clusters.

ke4:


upendrakumbham (approved):
LGTM!

PR #471, opened 2024-09-20
v16.0.0

ke4:
Start using versions for the app (v16.0.0).
Only the version of the web app has changed.
We can get the version by using the `/json/build` endpoint.


ke4:
As it is just a tiny change in a gradle properties file, just the version property has changed in that file, I am going to merge it. I don't think it needs an approval.

PR #483, opened 2024-10-30
Upgrade scxa-marker-gene-heatmap versions

### LINKED ISSUE #173: [Feature] Overlapping names in the marker gene heat map

lingyun1010:
I've tested locally with the new package version integrated, which works well. Besides, I tried to change the browser window size and it still behave well. Hence, I suggest to deploy to test environment first and check with other team members to see if they have any UI problems.
Screenshot 2024-10-30 at 16 42 03

Screenshot 2024-10-30 at 16 42 14


ke4 (approved):
LGTM
Let's deploy and test it!

PR #485, opened 2024-11-05
Bugfix eslint version mismatch

lingyun1010:
With the previous fix, it worked on my local machine ( I suspect that my local machine kept some build cache) but didn't pass the jenkins build on [develop branch](http://gene-expression.ebi.ac.uk/jenkins/job/Single%20Cell%20Expression%20Atlas%20%E2%80%93%20Develop/job/develop/), while the mysterious thing is Jenkins passed in my last PR.

It is due to Jenkins build node_modules for modules which has been deleted by me to avoid `eslint-config` mismatch version issues. So in this test, I recover that command but made some modifications. The idea is to filter our group's packages but ignore `eslint-config`, as not every packages are up to date with this one.

However, although we don't update `eslint-config` package, there sill are some peer dependency issues under `modules`, so we also need to specify the peer dependency explicitly to get rid of error messages. Please refer to [experiments-summary-panel](https://github.com/ebi-gene-expression-group/experiments-summary-panel/pull/4) and [species-summary-panel](https://github.com/ebi-gene-expression-group/species-summary-panel/pull/4).

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,7 +302,7 @@ For the first step you can run the following script:
  ./compile-front-end-packages.sh -iu
 ```
 
-The sceond step is simply:
+The second step is simply:
 ```bash
 ./gradlew :app:war
 ```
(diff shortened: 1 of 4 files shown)


ke4 (dismissed):
LGTM

PR #488, opened 2024-11-08
Fix ordering of the cluster labels in the marker gene heat map

ke4:
When the cluster labels in the marker gene heat map were numeric, their order was not correct.
For example it looked like this: 1, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 2, 20, 21, 3, 4, 5, 6, 7, 8, 9
The expected order is this one: 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21.

With the change of the comparator for the marker gene it is working fine now.


lingyun1010 (approved):
LGTM！

PR #497, opened 2024-12-19
Set HttpOnly argument for cookies true in the HTTP request headers

### LINKED ISSUE #496: Add HttpOnly header to our web pages
We have a cross-site scripting vulnerability on the www.ebi.ac.uk/gxa site.
This site has not got HttpOnly headers for cookies. Without this header these cookies are accessible through Javascript, that can be leveraged by a malicious actor to steal user session data and hijack accounts.

https://owasp.org/www-community/HttpOnly

ke4:
With this change we increase the security of the cookies by set the HttpOnly parameters of the cookies to true.
We set it in 3 different ways in the backend:

- in the web.xml file
- in the context.xml file
- create a new listener that set this argument (HttpOnly) to true.

We are going to do some changes in the frontend in another PR.


upendrakumbham (approved):
LGTM!

PR #505, opened 2025-03-28
Update download.jsp - replace Louvain with Leiden clustering 

pmb59:
To be merged once this is in implemented widely in production (e.g. after an Ensembl reference update).


irisdianauy (approved):
LGTM

anilthanki:
May be we need to think about if we are reanalysing all dataset or not, if not then old analysis are done with `louvain` and new with `Leiden`

PR #525, opened 2025-04-25
[Chore] #513 - Replace JSP views with Thymeleaf templates for experiments

### LINKED ISSUE #513: Transform experiment page to HTML/ThymeLeaf
This task should transform the experiment page from JSP to HTML/ThymeLeaf.

ke4:
Migrated experiment-page views from JSP to Thymeleaf to modernize the templating engine. Updated the controller, test cases, and resource files to align with the new Thymeleaf setup, ensuring consistent behavior and improved maintainability.


app/src/main/resources/templates/thymeleaf/views/experiment-page/experiment-description.html
```diff
@@ -0,0 +1,69 @@
+<div id="experimentDescription" class="row column expanded">
+    <div class="media-object stack-for-small" th:classappend="${#strings.startsWith(experimentAccession, 'E-ANND-')} ? 'anndata' : ''">
```
upendrakumbham, line 2: Is it possible to move hardcoded experiment accessions like 'E-ANND- ' to a constant file like Backend?

ke4: Not in this case.
`th:classappend` as its name tells append a string to the `class` property of the `div` element.
In this case, if `experimentAccession` starts with `E-ANND-` then the `class` property's value will be: `media-object stack-for-small anndata` otherwise its value stays `media-object stack-for-small`.
I am not sure what you mean by move it to the backend?
Thanks

upendrakumbham: ok..thanks for the clarification.

upendrakumbham (approved):
LGTM

PR #554, opened 2025-09-11
[Chore] GXA-339 Reorder featured experiment collections

ke4:
Adjusted the order of cards in the `featuredExperimentsCards` method to prioritize the "Human Cell Atlas" entry at the top. This ensures a consistent and intended display sequence for featured experiments.


ke4:
This is a tiny change. I have tested it in my local env and it worked fine.
We discussed it with the other devs that it does not require a review, so I am going to merge without it.

PR #559, opened 2026-03-31
scxa on k8s, ci GXA-418

amnonkhen:
Amongst the fixes in my branch:

optimisation of ui loading

1. safe loading of experiments on webapp startup
2. some local build scripts and config for vs code
3. a docker file (which we will not use)
4. optimisation of ui loading


sandsebi (approved):
Looks good! 👍

